PR #2, opened 2019-03-29
DO NOT MERGE: generated TypeScript code

alexander-fenster:
This code was generated by TypeScript micro-generator (resides in [proof-of-concept](https://github.com/googleapis/gapic-generator-typescript/tree/proof-of-concept) branch, PR pending - gosh, it will be a HUGE PR).

I actually verified it compiles and works against Showcase docker image. Before I go further and add LRO, paging, and streaming methods, please feel free to review the TypeScript part (imports and type definitions) and let me know if you have any questions!

For your reference: [here](https://20665-56533838-gh.circle-artifacts.com/0/tmp/workspace/gapic-generator/artman-genfiles/js/showcase-v1alpha3/src/v1alpha3/echo_client.js) is the code of the same library generated by gapic-generator.

TODO:
- LRO, paging, streaming: `_innerApiCalls` are ready but need to define methods with types.
- Generated tests
- Docker image and wrapper script (the one that puts protos in place and calls `protoc`)
- CI


package.json
```diff
@@ -0,0 +1,40 @@
...
+  "main": "build/src/index.js",
+  "files": [
+    "protos",
+    "build/src",
+    "!build/src/**/*.map",
+    "AUTHORS",
+    "COPYING"
+  ],
+  "dependencies": {
+    "google-gax": "googleapis/gax-nodejs"
```
alexander-fenster: I will release the new version when I'm sure no more changes are required.

alexander-fenster:
Sample JS code that uses this library (compatible with what we had before):

```js
const Client = require('./build/src/index');
const grpc = require('grpc');

async function main() {
  const echo = new Client.EchoClient({ 
  // also possible: const echo = new Client.v1alpha3.EchoClient({ 
    sslCreds: grpc.credentials.createInsecure(),
    servicePath: 'localhost',
    port: 7469
  });
  const response = await echo.echo({ content: 'test' });
  console.log(response);
}

main().catch(console.error);
```

Sample TS code that uses the library:
```ts
import * as Client from '.';
import {credentials} from 'grpc';
import * as protos from '../pbjs-genfiles/protos';
const ServiceTypes = protos.google.showcase.v1alpha3;

async function main() {
  const echo = new Client.v1alpha3.EchoClient({
  // also possible: const echo = new Client.v1alpha3.EchoClient({
    sslCreds: credentials.createInsecure(),
    servicePath: 'localhost',
    port: 7469
  });
  const response = await echo.echo(ServiceTypes.EchoRequest.create({content: 'test'}));
  console.log(response);
}

main().catch(console.error);
```

alexander-fenster:
@stephenplusplus @callmehiphop FYI as well. This is the first piece of generated TypeScript code (and sample JS and TS code that uses it). Now it's the best time for general feedback about typings and approaches.

package.json
```diff
@@ -0,0 +1,40 @@
...
+  "name": "showcase",
+  "version": "0.1.0",
+  "author": "Google LLC",
+  "description": "GAPIC Showcase client for Node.js",
+  "main": "build/src/index.js",
+  "files": [
+    "protos",
+    "build/src",
+    "!build/src/**/*.map",
+    "AUTHORS",
```
JustinBeckwith: we want to drop `AUTHORS` and `COPYING`

package.json
```diff
@@ -0,0 +1,40 @@
...
+    "COPYING"
+  ],
+  "dependencies": {
+    "google-gax": "googleapis/gax-nodejs"
+  },
+  "devDependencies": {
+    "@types/mocha": "^5.2.5",
+    "@types/through2": "^2.0.34",
```
JustinBeckwith: We really shouldn't need `through2` after node6 goes EOL - it's just a `PassThrough` stream.

src/v1alpha3/echo_client.ts
```diff
@@ -0,0 +1,233 @@
...
+  page: {[name: string]: gax.PageDescriptor};
+  stream: {[name: string]: gax.StreamDescriptor};
+  longrunning: {[name: string]: gax.LongrunningDescriptor};
+}
+
+export interface Callback<ResponseObject> {
+  (err: Error|null|undefined, value?: ResponseObject|null): void;
+}
+
+export interface PaginationCallback<
```
JustinBeckwith: Wild question time.  If this is a breaking change - how user hostile would it be to have an async first API, and not natively support callbacks? @bcoe @ofrobots

src/v1alpha3/echo_client.ts
```diff
@@ -0,0 +1,233 @@
...
+  clientConfig?: gax.ClientConfig;
+}
+
+interface Descriptors {
+  page: {[name: string]: gax.PageDescriptor};
+  stream: {[name: string]: gax.StreamDescriptor};
+  longrunning: {[name: string]: gax.LongrunningDescriptor};
+}
+
+export interface Callback<ResponseObject> {
```
JustinBeckwith: I would expect the `ResponseObject` type here to be generic

src/v1alpha3/echo_client.ts
```diff
@@ -0,0 +1,233 @@
...
+
+export interface PaginationResponse<
+    RequestObject, ResponseObject, ResponseType> {
+  values?: ResponseType[];
+  nextPageRequest?: RequestObject;
+  rawResponse?: ResponseObject;
+}
+
+export class EchoClient {
+  _descriptors: Descriptors = {page: {}, stream: {}, longrunning: {}};
```
JustinBeckwith: With the underscore, I'd expect this to be `private`

src/v1alpha3/echo_client.ts
```diff
@@ -0,0 +1,233 @@
...
+  _innerApiCalls: {[name: string]: Function};
+  auth: gax.GoogleAuth;
+
+  constructor(opts: ClientOptions) {
```
JustinBeckwith: Just to have the discussion - are we dropping traditional jsdoc annotations?

src/v1alpha3/echo_client.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    // Save the auth object to the client, for use by other methods.
+    this.auth = gaxGrpc.auth;
+
+    // Determine the client header string.
+    const clientHeader = [
+      `gl-node/${process.version}`,
+      `grpc/${gaxGrpc.grpcVersion}`,
+      `gax/${gax.version}`,
+      `gapic/${version}`,
```
JustinBeckwith: Can we start to include the version of node.js being used with this data?

callmehiphop:
Personally I'm not a huge fan of making TS users create proto structs, I think `pbts` also creates an interface version for each of these objects (`EchoRequest` -> `IEchoRequest`) that can be used to represent plain objects. 

On that note, I'm actually not a fan of using `pbjs` and `pbts` either, we've had some issues with it pulling in global types. Plus if we were to use just the interfaces, then majority of those files do nothing but take up space. I'm not trying to advocate re-inventing the wheel here, but it does seem as though we have adequate data on the request/response objects that we may be able to generate them pretty easily.

src/v1alpha3/echo_client.ts
```diff
@@ -0,0 +1,233 @@
...
+    const echoStub =
+        // @ts-ignore We cannot check types that are loaded in runtime.
+        gaxGrpc.createStub(gaxProtos.google.showcase.v1alpha3.Echo, opts);
+
+    const echoStubMethods =
+        ['echo', 'expand', 'collect', 'chat', 'pagedExpand', 'wait'];
+
+    for (const methodName of echoStubMethods) {
+      this._innerApiCalls[methodName] = gax.createApiCall(
+          echoStub.then(
```
JustinBeckwith: This is a little funky.  What's with the `then` inside of a constructor?

src/v1alpha3/echo_client.ts
```diff
@@ -0,0 +1,233 @@
...
+              // @ts-ignore
+              stub => function() {
+                const args = Array.prototype.slice.call(arguments, 0);
```
JustinBeckwith: lets use `...args`, and drop the need to do the funky slice

src/v1alpha3/echo_client.ts
alexander-fenster: I would love doing that (and removing tons of overloads in the bottom of that file), but it obviously would be a kind of breaking change that is indeed breaking. To be discussed.

src/v1alpha3/echo_client.ts
alexander-fenster: Fixed!

src/v1alpha3/echo_client.ts
alexander-fenster: I didn't bother adding annotations to make the code readable while we're in this phase. I'll need to add _some_ annotations anyway.

src/v1alpha3/echo_client.ts
alexander-fenster: Isn't it in the first header? `${process.version}` :)

src/v1alpha3/echo_client.ts
alexander-fenster: No changes compared with the current code (example here: https://github.com/googleapis/nodejs-cloud-container/blob/0d3ba60b9ec6b7f0b6debe9b6c5d653260b3d9d2/src/v1/cluster_manager_client.js#L154)

The reason of having `.then` in constructor is that async authentication happens in `gaxGrpc.createStub`, so we either have delayed method initialization here, or must have `async auth()` in the client class, or `async` factory function. The two latter options would be really breaking so I'd prefer to leave it as it is now, even though it's really not obvious (I always look at this code for a few minutes before I realize how it works, every time).

src/v1alpha3/echo_client.ts
alexander-fenster: Fixed! Thank you for this suggestion!

package.json
alexander-fenster: I will try removing it and will see what breaks :)

src/v1alpha3/echo_client.ts
alexander-fenster: I didn't quite understand you here - can you explain please? I was under impression it _is_ generic now, since it's `Callback<ResponseObject>`.

src/v1alpha3/echo_client.ts
JustinBeckwith: I am not a smart man

alexander-fenster:
@callmehiphop Thank you for the tip about interfaces! I updated the generator to use them, the TS sample is now much simpler:

```ts
import {credentials} from 'grpc';

import * as Client from '.';

async function main() {
  const echo = new Client.EchoClient({
    // also possible: const echo = new Client.v1alpha3.EchoClient({
    sslCreds: credentials.createInsecure(),
    servicePath: 'localhost',
    port: 7469
  });
  const response = await echo.echo({content: 'test'});
  console.log(response);
}

main().catch(console.error);
```

I don't really want to recreate type definitions manually given that we have them generated for free by `pbts` but we can discuss our options.

src/v1alpha3/echo_client.ts
JustinBeckwith: Where does the `ResponseObject` type definition come from?

callmehiphop: I think `ResponseObject` is the type variable, so it shouldn't be defined anywhere

alexander-fenster: @JustinBeckwith It's substituted with an actual type for each auto-generated method: 
```
  echo(
      request: protos.google.showcase.v1alpha3.IEchoRequest,
      options: gax.CallOptions,
      callback: Callback<protos.google.showcase.v1alpha3.IEchoResponse>): void;
```

PR #4, opened 2019-08-07
Call method without arguments or empty object

xiaozhenliu-gg5:
* Allow method calling without required argument for non-streaming and server-streaming methods


.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+
```
alexander-fenster: We normally don't checkin editor-specific files. You can add the whole `.idea` folder to `.gitignore`.

alexander-fenster: The `.DS_Store` files should not be checked in either.

alexander-fenster:
`.DS_Store` files are still there, can you please delete them as well? Thank you!

.idea/.gitignore
xiaozhenliu-gg5: Sure, done.

PR #5, opened 2019-08-13
test: add baseline test

xiaozhenliu-gg5:
1. create showcase directory
2. create showcase/tmp directory 
3. run protoc command line to generate client code
4. compare echo_client results

TODO: 
Use google protos from release page, instead of local path.


typescript/test/CodeGeneratorTest.ts
```diff
@@ -0,0 +1,50 @@
+const execa = require('execa');
+const util = require('util');
+const path = require('path');
+const fs = require('fs');
+
+const SHOWCASE_CLIENT_LIB = './showcase';
+const TMP_CLIENT_LIB = './showcase/tmp';
```
alexander-fenster: Let's use `path.join` everywhere to make it compatible across different OS (Windows is a real use case). `path.join(cwd, 'showcase', 'tmp');` where `cwd = process.cwd()` - here and below. There should be no forward slashes anywhere in the code (Windows uses backslashes).

.gitignore
```diff
@@ -14,3 +14,4 @@ package-lock.json
 pbjs-genfiles/
 .DS_Store
 .idea
+showcase/
```
alexander-fenster: nit: newline

typescript/test/CodeGeneratorTest.ts
```diff
@@ -0,0 +1,50 @@
...
+//  download will give the whole repo.
+const LOCAL_CLIENT_LIB_DIR = '-I/Users/xiaozhenliu/showcase';
+const PROTO_DIR = 'google/showcase/v1beta1/echo.proto';
+const CLIENT_LIBRARY_BASELINE = path.join(__dirname, '..', '..') +
+    '/typescript/test/testdata/echo_client_baseline';
+
+describe('CodeGeneratorTest', () => {
+  describe('Create ./showcase dir and get the protos', () => {
+    it('Download showcase repo in ./showcase dir.', () => {
+      // Create ./showcase directory.
```
alexander-fenster: Maybe just `fs.mkdir`?

typescript/test/CodeGeneratorTest.ts
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  describe('Generate client library', () => {
+    it('Generated client library should have same output with baseline.',
+       async () => {
+         await execa('protoc', [
+           GENERATED_CLIENT_LIB_DIR, GOOGLE_GAX_PROTOS_DIR,
+           LOCAL_CLIENT_LIB_DIR, PROTO_DIR
+         ]);
+         await execa('cmp', [GENERATED_CLIENT_FILE, CLIENT_LIBRARY_BASELINE]);
```
alexander-fenster: Let's minimize external dependencies and compare files manually. `cmp` might not be available everywhere. Just make your own `function cmp(file1, file2) { }` that uses `fs.readFileSync`.

alexander-fenster:
Thank you, that's awesome! Let's make it portable (I left some minor comments) and then I'll enable CI for this repository so that this test runs for all checkins. After that we'll be able to move our development to master branch.  🎉

typescript/test/CodeGeneratorTest.ts
xiaozhenliu-gg5: Thanks for the comments! 
I have a question about writing the customize compare function. It complains about Syntax Error:Unexpected token : at the compareFlies(file1: [<- here] string, file2: string){} Confused about why.

xiaozhenliu-gg5:
I'm using the local showcase protos in the test, ideally we should download it somewhere or hardcode some g3 path for stable testing. But I'm not sure where is the best place to get them [LOCAL_CLIENT_LIB_DIR = /Users/xiaozhenliu/showcase/google/showcase/echo.proto]?

typescript/test/CodeGeneratorTest.ts
```diff
@@ -0,0 +1,36 @@
...
+           fs.mkdirSync(SHOWCASE_CLIENT_LIB);
+         }
+         if (!fs.existsSync(TMP_CLIENT_LIB)) {
+           fs.mkdirSync(TMP_CLIENT_LIB);
+         }
+         await execa('protoc', [
+           GENERATED_CLIENT_LIB_DIR, GOOGLE_GAX_PROTOS_DIR,
+           LOCAL_CLIENT_LIB_DIR, PROTO_DIR
+         ]);
+         assert(fs.readFileSync(GENERATED_CLIENT_FILE).equals(fs.readFileSync(CLIENT_LIBRARY_BASELINE)));
```
alexander-fenster: This will work for now. Please run `npx gts fix` to format the code :)

typescript/test/CodeGeneratorTest.ts
alexander-fenster: That's weird, but I see you worked around that problem by calling `readFileSync` directly. That will work for now.

alexander-fenster:
Re: showcase protos - we can either download them from Github (not from g3) or commit them into the test folder. I think it's OK to commit the protos to make the test hermetic.

typescript/test/CodeGeneratorTest.ts
xiaozhenliu-gg5: Yes, using the function directly solves the problem

xiaozhenliu-gg5:
Adding showcase protos to test folder sounds a great idea.

alexander-fenster:
I'm making a change to have `protoc-gen-typescript_gapic` in the test environment - still working on it. I'll update this PR when ready (hopefully later today).

typescript/test/CodeGeneratorTest.ts
```diff
@@ -0,0 +1,42 @@
+const execa = require('execa');
```
JustinBeckwith: Can we use the `import` syntax here?

typescript/test/CodeGeneratorTest.ts
```diff
@@ -0,0 +1,42 @@
...
+  describe('Generate client library', () => {
+    it('Generated client library should have same output with baseline.',
+       async () => {
+         if (!fs.existsSync(SHOWCASE_CLIENT_LIB)) {
+           fs.mkdirSync(SHOWCASE_CLIENT_LIB);
+         }
+         if (!fs.existsSync(TMP_CLIENT_LIB)) {
+           fs.mkdirSync(TMP_CLIENT_LIB);
+         }
+         await execa('protoc', [
```
JustinBeckwith: You probably don't need this to be async, and you probably don't need execa :)

```js
const {execSync} = require('child_process');
execSync(`protoc ${GENERATED_CLIENT_LIB_DIR}  ${GOOGLE_GAX_PROTOS_DIR, ${LOCAL_CLIENT_LIB_DIR, PROTO_DIR}`);
```

or something like that should be fine

package.json
```diff
@@ -52,6 +52,7 @@
   },
   "dependencies": {
     "command-line-args": "^5.0.2",
+    "execa": "^2.0.3",
```
JustinBeckwith: Per below, not sure we really need that

typescript/test/CodeGeneratorTest.ts
```diff
@@ -0,0 +1,42 @@
...
+           GENERATED_CLIENT_LIB_DIR, GOOGLE_GAX_PROTOS_DIR,
+           LOCAL_CLIENT_LIB_DIR, PROTO_DIR
+         ]);
+         assert(fs.readFileSync(GENERATED_CLIENT_FILE)
```
JustinBeckwith: This is a weird way to do the assertion.  Can we use `assert.strictEquals` instead?

typescript/test/testdata/echo_client_baseline
```diff
@@ -0,0 +1,234 @@
+// Copyright 2019 Google LLC
```
JustinBeckwith: This isn't blocking feedback.  For these baselines - should we consider using something similar to the snapshotting model @bcoe is using in other cmdline apps?

bcoe: here's an example of using snapshots to test a CLI:

https://github.com/bcoe/c8/blob/master/test/integration.js#L18

It can be a really nice pattern if you have large payloads you're asserting against, updating tests becomes more visual. Where, if you make a change, and tests start failing, you eyeball the snapshot to make sure that the change in formatting was expected, if it was you simply run `test:snap`, to recapture snapshots.

What's nice is you automate the process of maintaining large fixtures.

alexander-fenster: I suggest we merge this one as is just because I want to have *some* baseline test ASAP, and then change it to a better engineered version. For now, I just need to be confident when making change, so I asked @xiaozhenliu-gg5 to add the very basic baseline test.

typescript/test/CodeGeneratorTest.ts
xiaozhenliu-gg5: Done. Thanks

package.json
xiaozhenliu-gg5: Done.

typescript/test/CodeGeneratorTest.ts
xiaozhenliu-gg5: I tried assert.strictEqual(buffer1, buffer2) but it will give error even if I copy paste the generated client library as baseline result. Not sure the best way to compare two files content.

alexander-fenster: Resolved locally - `assert.strictEqual(buffer1.toString(), buffer2.toString())`

PR #24, opened 2019-09-19
feat: add proto list template

xiaozhenliu-gg5:
One way with helper function to parse proto path.


templates/typescript_gapic/src/$version/$service_proto_list.json.njk
```diff
@@ -0,0 +1,3 @@
+[
+  {{ api.protoFilesToGenerate | safe}}
+]
```
alexander-fenster: Please add an empty line

typescript/src/schema/api.ts
```diff
@@ -52,4 +52,11 @@ export class API {
       proto => this.protos[proto].fileToGenerate
     );
   }
+
+  get protoFilesToGenerate(){
+    return this.filesToGenerate.map(function(file){
+      return '\"../../protos/' + file + '\"' + ',\n';
```
alexander-fenster: Please use `\`"../../protos/${file}"` instead of concatenation. Also, it makes sense to use `.join` for `',\n'` in the template (and not here), because commas and newlines are presentation and belong to a template.

alexander-fenster:
`gts` requires some minor changes :)

alexander-fenster:
Also, maybe it's a time to get a better baseline coverage.

typescript/src/generator.ts
```diff
@@ -17,9 +17,15 @@ import * as path from 'path';
 
 import * as plugin from '../../pbjs-genfiles/plugin';
 
+<<<<<<< HEAD
```
alexander-fenster: This looks very wrong :)

typescript/test/baseline.ts
```diff
@@ -54,6 +54,21 @@ const PLUGIN = path.join(SRCDIR, 'protoc-gen-typescript_gapic');
 
 describe('CodeGeneratorTest', () => {
   describe('Generate client library', () => {
+<<<<<<< HEAD
```
alexander-fenster: This too.

PR #33, opened 2019-10-03
test: showcase integration test set up

xiaozhenliu-gg5:
1. config.yml: 
- compile client library and install it to test-application && run script

2. bash script: 
- download showcase server and run it in background 
- run mocha test for js users
- terminate the server process


typescript/test_application/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "google-gax": "^1.6.0",
+    "gts": "^1.0.0",
+    "intelli-espower-loader": "^1.0.1",
+    "mocha": "^6.0.0",
+    "power-assert": "^1.6.0",
+    "rimraf": "^3.0.0",
+    "typescript": "~3.6.0"
+  },
+  "dependencies": {
+    "command-line-args": "^5.0.2",
```
alexander-fenster: I'm kind of sure the test application does not need any of these dependencies. It will be cleaner if it depends on `"showcase": "file:///../../path/to/showcase/client"`, then you will be able to do 

```js
const showcase = require('showcase');
```

and it will use the client library exactly as the real users will do it.

typescript/test_application/showcase.js
```diff
@@ -0,0 +1,178 @@
...
+          };
+        }
+      };
+      const fallbackClientOpts = {
+        fallback: true,
+        protocol: 'http',
+        port: 1337,
+        auth: fakeGoogleAuth,
+      };
+      assert.deepStrictEqual('1', '2');
```
alexander-fenster: This should fail?

typescript/test_application/showcase-setup.sh
```diff
@@ -0,0 +1,23 @@
+mkdir showcase-server 
+cd showcase-server
+export GAPIC_SHOWCASE_VERSION=0.5.0
+export OS=darwin
```
alexander-fenster: Won't work in Circle :(

alexander-fenster:
Thanks @xiaozhenliu-gg5, that's great!

Right now it does not really work even though it's green in Circle :)

The server cannot be started since the script downloads it for `darwin` but Circle runs on Linux. So start fails. But for some reason the tests don't really fail :)  so you get a green build.  It actually has errors:

image

The whole setup looks good to me, let's make a few changes though:
- let the test app depend on `showcase` client library via `package.json` dependency;
- let's use exactly the same test app as in `gapic-generator`, which will give us more confidence that the libraries are compatible - I suggest reusing [this one](https://github.com/googleapis/gapic-generator/tree/master/showcase/nodejs) with zero changes and it must work, otherwise we won't be able to replace gapic-generated client with micro-generator-generated client!
- when you make it work with the shell script, I suggest replacing the shell script with a JavaScript or TypeScript file to make the test runnable on Windows.

Does it make sense?

As a nit, it might make sense to move the `test_application` somewhere inside the `test` folder.

alexander-fenster:
Also, this is a test change, so it's a `test:`, not a `feat:` :)

alexander-fenster:
Explanation: `feat:` is something that bumps semver minor. `fix` is something that bumps semver patch version. Anything with `!` bumps semver major. Test changes are `test:`, documentation changes are `docs:`, test changes are `test:`, random stuff of no interest for users is `chore:`.

More information: https://www.conventionalcommits.org/en/v1.0.0-beta.4/

xiaozhenliu-gg5:
Thanks for detailed comments Alex! It's very helpful, I will make the script work well in circleCI first and then move the test application, make it same as in gapic-generator.

But I have question referring to move the test-application into test folder, since in root dir, when we run `npm test`, it will run all the tests in test folder, correct, which is `mocha build/test`? That will include our test-application, and the test-application now actually uses generated client library which is created by `npm test`, that would be a circle.  Please give some advice here. Thank you!

xiaozhenliu-gg5:
Thanks for the suggestion:
1. change `env` to linux make downloading server correctly
2. update showcase-test to test `async` function and detect when it's done
Now it fails for `assert.equal('1', '2')` and passes for our tests.
3. clean package.json dependencies and use real client lib

alexander-fenster:
> But I have question referring to move the test-application into test folder, since in root dir, when we run npm test, it will run all the tests in test folder, correct, which is mocha build/test

At least two options here:
1. Use `mocha build/test/*.js` for the test command to limit it to `*.js` files in that one folder
2. Move tests to `typescript/test/unit`, make the command `mocha build/test/unit` (like we do in `google-gax`).

alexander-fenster:
OK now it actually works in Circle! 🎉 Now let's get rid of shell in the pipeline and switch to the same test app as in gapic-generator and we'll be good to go.

xiaozhenliu-gg5:
Sure, got it! Thanks!

.gitignore
```diff
@@ -16,3 +16,4 @@ pbjs-genfiles/
 .idea
 .baseline-test-out
 .multi-proto-test-out
+typescript/test_application/showcase-server
```
alexander-fenster: This path is now incorrect since the application has moved

typescript/test/test_application/karma.conf.js
```diff
@@ -0,0 +1,102 @@
+/**
```
alexander-fenster: I'm not checking this file, it might needs to be changed when we check the browser use case but for now let's just commit it as is.

typescript/test/test_application/showcase-setup.sh
```diff
@@ -0,0 +1,22 @@
+mkdir showcase-server 
+cd showcase-server
+export GAPIC_SHOWCASE_VERSION=0.5.0
+export OS=`uname`
```
alexander-fenster: Let's do `uname | tr A-Z a-z` to make sure it's lowercase. On my Mac, `uname` gives `Darwin` with the capital D (but if you verified it works, we can leave it as is)

typescript/test/test_application/showcase-setup.sh
```diff
@@ -0,0 +1,22 @@
...
+export ARCH=amd64
```
alexander-fenster: `uname -m`

alexander-fenster: Well... nope, let's leave `amd64`. We don't publish binaries for other platforms anyway.

package.json
```diff
@@ -13,7 +13,7 @@
     "protoc-gen-typescript_gapic": "build/src/cli.js"
   },
   "scripts": {
-    "test": "c8 --reporter=lcov mocha build/test",
+    "test": "c8 --reporter=lcov mocha build/test/unit",
     "codecov": "c8 --reporter=lcov mocha build/test && c8 report",
```
alexander-fenster: Also `build/test/unit` here.

typescript/test/test_application/karma.conf.js
xiaozhenliu-gg5: Left a TODO comment to track the undone work.

typescript/test/test_application/showcase-setup.sh
xiaozhenliu-gg5: Confirmed all tests pass on my side and linux OS, thanks!

xiaozhenliu-gg5:
Thanks a lot for all the feedback Alex! :)

PR #42, opened 2019-10-10
test: add long running method in test application for ts users

xiaozhenliu-gg5:
1. expose longrunning response type & metadata type for method.
2. add long running method tests.
3. update baseline test.


typescript/src/schema/proto.ts
```diff
@@ -106,11 +132,17 @@ function toInterface(type: string) {
   return type.replace(/\.([^.]+)$/, '.I$1');
 }
 
+function toLRInterface(type: string, inputType: string) {
+  return inputType.substring(1, inputType.lastIndexOf('.') + 1) + 'I' + type;
```
alexander-fenster: This deserves a comment. Even better, I strongly prefer making this a regex replace since it's often much more readable.

PR #43, opened 2019-10-10
test: add paged method to test application for ts users

xiaozhenliu-gg5:


typescript/test/test_application_ts/src/index.ts
```diff
@@ -103,3 +104,18 @@ async function testCollect(client: showcase.v1beta1.EchoClient) {
...
+async function testPagedExpand(client: showcase.v1beta1.EchoClient) {
+  const words = ['nobody', 'ever', 'reads', 'test', 'input'];
+  const request = {
+    content: words.join(' '),
+    pageSize: 2,
+  };
+  const [response] = await client.pagedExpand(request);
+  const result = response as Array<JSON>;
+  const expectedResponse = [];
+  for(var i = 0; i < words.length; i++){
```
alexander-fenster: `expectedResponse = words.map(word => { content: word });`

typescript/test/test_application_ts/src/index.ts
```diff
@@ -103,3 +104,18 @@ async function testCollect(client: showcase.v1beta1.EchoClient) {
...
 }
+
+async function testPagedExpand(client: showcase.v1beta1.EchoClient) {
+  const words = ['nobody', 'ever', 'reads', 'test', 'input'];
+  const request = {
+    content: words.join(' '),
+    pageSize: 2,
+  };
+  const [response] = await client.pagedExpand(request);
+  const result = response as Array<JSON>;
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: yes, response is like [{content: '1'}, {content: '2'}...], which is `IpagedResponseType`. Technically it should be response.map(r=> r.content) while map property is not supported for the pagedResponseType

typescript/test/test_application_ts/src/index.ts
xiaozhenliu-gg5: This is a better way but tsc compile will complain about the unused label `content`

xiaozhenliu-gg5: But I can `"allowUnusedLabels": true,` to the tsconfig file if that's ok

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -400,29 +400,29 @@ export class {{ service.name }}Client {
       request: protosTypes{{ method.inputInterface }},
       options?: gax.CallOptions):
       Promise<[
-        protosTypes{{ method.outputInterface }},
+        protosTypes{{ method.pagingResponseType }}[],
         protosTypes{{ method.inputInterface }}|undefined, 
         {}|undefined
```
alexander-fenster: OK I checked in gax: it calls `callback(err, resources, request, response)` or `callback(err, resources, null, response)`, so the correct types here would be:
- first parameter: `protosTypes{{ method.pagingResponseType }}[]`  (it's correct here)
- second parameter: request type or `null` (no `undefined` from what I see), so please change `undefined` to `null`
- third parameter: actual response, so it should be the response type (`IPagedExpandResponse`). I don't think it can be `undefined`.

typescript/src/schema/proto.ts
```diff
@@ -99,7 +100,25 @@ function pagingFieldName(messages: MessagesMap, method: MethodDescriptorProto) {
...
+}
+
+function pagingResponseType(
+  messages: MessagesMap,
+  method: MethodDescriptorProto
+) {
+  const repeatedFields = pagingField(messages, method);
+  if (repeatedFields && repeatedFields.typeName) {
+    const typeName = repeatedFields.typeName; //.google.showcase.v1beta1.EchoResponse
+    return typeName.replace(/.([^.]*)$/, '.I$1');
```
alexander-fenster: A dot is a special character in regexes and must be written as \. if it's not within a character class. Outside of a character class, . means 'any character'. Your regex works because it's applied from left to right and it turns out that the final dot is the first character that starts the first possible match, but still, let's make it /\.([^.]+)$/ (+ instead of * just in case).

Issue #18, opened 2019-08-30
generate unit tests


alexander-fenster:
We need to generate unit tests with 100% code coverage of the generated library code.

alexander-fenster:
Fixed by #46, #38, #37

PR #46, opened 2019-10-14
test: add paging method in unit test

xiaozhenliu-gg5:
Add paging method unit test for generated client library.

TODO: init request and response based on structure function, now the test only uses empty request to trigger the method and get empty response back.
#18


alexander-fenster:
Also, I just realized our unit test is `.js`. It should be `.ts`. I'll submit an issue :)

alexander-fenster:
#48!

xiaozhenliu-gg5:
Got it next I will expand the baseline test! and also convert the js to ts :)

PR #49, opened 2019-10-14
feat: add starter script for micro-gen

### LINKED ISSUE #45: write starter script for micro-gen

xiaozhenliu-gg5:
test locally with two commands below:
1. `gapic-generator-typescript --PROTOS_DIR=build/test/protos/ --PROTO_FILE=build/test/protos/google/showcase/v1beta1/echo.proto`
2. `gapic-generator-typescript --PROTOS_DIR=build/test/protos/ --PROTO_FILE=build/test/protos/google/showcase/v1beta1/echo.proto --OUTPUT_DIR=/Users/xiaozhenliu/gapic-generator-typescript/.baseline-test-out/`

The required args are `PROTOS_DIR` & `PROTO_FILE`, `OUTPUT_DIR` has a default value.
Fixes #45 

3. copy `protos` to generated client library & skip common gax protos


package.json
```diff
@@ -10,7 +10,8 @@
     "pbjs-genfiles"
   ],
   "bin": {
-    "protoc-gen-typescript_gapic": "build/src/cli.js"
+    "protoc-gen-typescript_gapic": "build/src/cli.js",
+    "gapic-geenrator-typescript": "build/src/start_script.js"
```
alexander-fenster: `generator`

typescript/src/start_script.ts
```diff
@@ -0,0 +1,76 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { execSync } from 'child_process';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as rimraf from 'rimraf';
```
alexander-fenster: this is scary - I would probably prefer to fail if the output directory exists and is not empty rather than delete the contents (even though `artman` would delete the contents)

typescript/src/start_script.ts
```diff
@@ -0,0 +1,76 @@
...
+import * as readline from 'readline';
+const argv = require('yargs').argv;
```
alexander-fenster: this does not look like a proper TypeScript, is it possible to `import` it properly?

typescript/src/start_script.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const cwd = process.cwd();
+
+const GOOGLE_GAX_PROTOS_DIR = path.join(
+  cwd,
```
alexander-fenster: There is no guarantee that the `cwd` will contain `node_modules`. I would say in most cases the generator will be run from some other place. You can use `__dirname` as your starting point to locate the protos.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,76 @@
...
+  'node_modules',
+  'google-gax',
+  'protos'
+);
+const SRCDIR = path.join(cwd, 'build', 'src');
+const CLI = path.join(SRCDIR, 'cli.js');
+const PLUGIN = path.join(SRCDIR, 'protoc-gen-typescript_gapic');
```
alexander-fenster: I think the `PLUGIN` is actually in the `bin` directory, not in `SRCDIR`. Let's figure out tomorrow.
The logic to copy `PLUGIN` below should not be here, since we expect that the generator is installed locally. 

There are two options:
1. The generator is installed globally with `npm install -g gapic-generator-typescript`, then `protoc-gen-typescript_gapic` is in global Node.js `bin` directory.
2. The generator is installed locally with `npm install --save-dev gapic-generator-typescript`, then `protoc-gen-typescript_gapic` is in `./node_modules/.bin`.

In both cases, I'm guessing we can take `__dirname` and guess the right location of `protoc-gen-typescript_gapic`, and add it into `PATH`. We don't need to create this file manually anyway, since `npm` will do it for us.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,76 @@
...
+} else {
+  console.error('Proto file directory is not assigned.');
+}
+if (fs.existsSync(PLUGIN)) {
+  rimraf.sync(PLUGIN);
+}
+fs.copyFileSync(CLI, PLUGIN);
+process.env['PATH'] = SRCDIR + path.delimiter + process.env['PATH'];
+try {
+  execSync(`chmod +x ${PLUGIN}`);
```
alexander-fenster: `npm` will do it for us since we expect to have a properly installed package.

alexander-fenster:
That's a great first iteration @xiaozhenliu-gg5! Let's talk tomorrow about the two use cases we want to support (the code will be simplified since we likely don't need to create `PLUGIN` manually, `npm` will do it for us).

Also, let's have this script covered by a test (we can change the showcase test to properly install the generator package globally with `npm install -g .` and use this wrapper).

xiaozhenliu-gg5:
Thanks Alex! Fixed for some feedback and let's have discussion tomorrow for `PLUGIN` issue :)

.circleci/config.yml
```diff
@@ -17,6 +17,7 @@ jobs:
       - run:
           name: Run baseline tests
           command: |
+            sudo npm link
```
alexander-fenster: why do we need it?

typescript/src/start_script.ts
```diff
@@ -0,0 +1,62 @@
...
+const GOOGLE_GAX_PROTOS_DIR = path.join(
+  __dirname,
+  '..',
+  '..',
+  'node_modules',
+  'google-gax',
+  'protos'
+);
+
+// Add executable of plugin to PATH
```
alexander-fenster: nit: not executable but its folder

typescript/src/start_script.ts
```diff
@@ -0,0 +1,62 @@
...
+process.env['PATH'] = __dirname + path.delimiter + process.env['PATH'];
+
+let OUTPUT_DIR = '';
+if (argv.output_dir) {
+  OUTPUT_DIR = argv.output_dir as string;
+} else {
```
alexander-fenster: nit: empty else.
nit: `OUTPUT_DIR`, `PROTO_DIRS` are not constants anymore, let's lowercase them

typescript/src/start_script.ts
```diff
@@ -0,0 +1,62 @@
...
+  '..',
+  'node_modules',
+  'google-gax',
+  'protos'
+);
+
+// Add executable of plugin to PATH
+process.env['PATH'] = __dirname + path.delimiter + process.env['PATH'];
+
+let OUTPUT_DIR = '';
```
alexander-fenster: just leave it `undefined` (`let OUTPUT_DIR: string;`)

typescript/src/start_script.ts
```diff
@@ -0,0 +1,62 @@
...
+
+let PROTO_DIRS = [];
+if (argv.I) {
+  if (Array.isArray(argv.I)) {
+    PROTO_DIRS.push(...argv.I);
+  } else {
+    PROTO_DIRS.push(argv.I);
+  }
+}
+PROTO_DIRS = PROTO_DIRS.map(dir => `-I${dir}`);
```
alexander-fenster: I just realized this won't work if `dir` contains spaces. `"-I${dir}"` will fix it.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,62 @@
...
+const PROTO_FILES = [];
+if (Array.isArray(argv._)) {
+  PROTO_FILES.push(...argv._);
+} else {
+  PROTO_FILES.push(argv._);
+}
+
+execSync(
+  `protoc ` +
+    `-I${GOOGLE_GAX_PROTOS_DIR} ` +
```
alexander-fenster: Let's add quotes around all parameters since they can have spaces.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+PROTO_DIRS = PROTO_DIRS.map(dir => `-I${dir}`);
+
+const PROTO_FILES = [];
+if (Array.isArray(argv._)) {
+  PROTO_FILES.push(...argv._);
+} else {
+  PROTO_FILES.push(argv._);
+}
```
alexander-fenster: same here: if a filename contains spaces the command line will be broken. ````map(file => `"${file}"`)```` helps.

Alternatively, you can use `execFileSync` instead of `execSync` and pass command line parameters as an array. Then you won't need to deal with spaces.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,62 @@
...
+
+execSync(
+  `protoc ` +
+    `-I${GOOGLE_GAX_PROTOS_DIR} ` +
+    `${PROTO_DIRS} ` +
+    `${PROTO_FILES} ` +
+    `--typescript_gapic_out=${OUTPUT_DIR} `
```
alexander-fenster: We should fail if `OUTPUT_DIR` is undefined.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,62 @@
...
+);
```
alexander-fenster: We should fail if `execSync` returned with error (non-zero exit code).

alexander-fenster:
Thank you! I think it works, I only have some minor comments. Also, let's make sure we return non-zero exit code if `protoc` fails.

alexander-fenster:
I just realized...

We need one more thing!

One of the main reasons why we need this starter script is that we want someone to copy all the protos for us. The generator emits the list of proto files, now, after `protoc` is executed, we need to copy all those proto files into the `protos` folder of the generated code.

typescript/src/start_script.ts
xiaozhenliu-gg5: Done

typescript/src/start_script.ts
xiaozhenliu-gg5: Tried to use quotes instead of back stick, seems back stick supports the template string, for "I${file}" it returns `I${file}` not inputing the real file.

.circleci/config.yml
xiaozhenliu-gg5: this links the bin file to the js, without this circleCI complains about not finding `gapic-generator-typescript`. But not sure if there's other way around.

typescript/src/start_script.ts
xiaozhenliu-gg5: Sorry please ignore my previous comment, I misunderstood the suggestion.

.circleci/config.yml
alexander-fenster: OK, that sounds good! I did not know `npm link` does that.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,84 @@
...
+
+// Add folder of plugin to PATH
+process.env['PATH'] = __dirname + path.delimiter + process.env['PATH'];
+
+let outputDir = '';
+if (argv.output_dir) {
+  outputDir = argv.output_dir as string;
+} else {
+  process.exit(1);
```
alexander-fenster: We are doing great for robots (`process.exit(1)`) but not too good for humans (no error message) :)  Let's add `console.error('Output directory is required: --output_dir path');` before exiting.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// create protos folder to copy proto file
+const COPY_PROTO_DIR = path.join(outputDir, 'protos');
+if (!fs.existsSync(COPY_PROTO_DIR)) {
+  fs.mkdirSync(COPY_PROTO_DIR);
+}
+// copy proto file to generated folder
+try {
+  protoDirs.forEach(dir => {
```
alexander-fenster: That's a little bit more complicated than this. The generator dumps the required protos into `proto.list`, so we need to copy all of those. We don't want to copy extra files that are in `protoDirs`, just the required ones.

Note that `proto.list` has both gax common protos and protos from the library. Gax common protos don't need to be copied. We can talk tomorrow if this logic is unclear (it actually _is_ unclear).

typescript/src/start_script.ts
xiaozhenliu-gg5: Sounds good to me, it makes more sense to give error message before exit.

typescript/src/start_script.ts
xiaozhenliu-gg5: I just copied the .../protos right now, but it definitely needs all the protos (was hesitating about gax protos and proto files like echo.proto) For my understanding, we need loop the protos in proto.list, except the gax common protos. I will confirm with you tomorrow if I'm doing wrong changes, thanks for the comment Alex!

xiaozhenliu-gg5:
TODO: command parameters should be wrapped in quotes to ensure spaces in them don't break the command line. [while "param" will exactly render in command which will fail `protoc` command eg: `protoc  "-I../../../protos", "-I../../node_modules/google-gax"....]`

typescript/src/start_script.ts
alexander-fenster: This is not yet resolved. We either need quotes around each parameter, or use `execFileSync` instead of `execSync`.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+// copy proto file to generated folder
+try {
+  const PROTO_LIST = path.join(outputDir, 'proto.list');
+  const protos = fs
+    .readFileSync(PROTO_LIST)
+    .toString()
+    .split('\n');
+  // skip common gax proto files
+  const requiredProtos: string[] = [];
```
alexander-fenster: Let's go functional here!
```
    .split('\n')
    .filter(proto => !fs.existsSync(.....))
    .forEach(proto =>  /* copy it here */ );
```

typescript/src/start_script.ts
```diff
@@ -0,0 +1,104 @@
...
+  requiredProtos.forEach(proto => {
+    protoDirs.forEach(dir => {
+      const protoFile = path.join(dir, proto);
+      if (fs.existsSync(protoFile)) {
+        fileSystem.copyFileSync(protoFile, path.join(COPY_PROTO_DIR, proto));
+      }
+    });
+  });
+} catch (err) {
+  console.error('Copy proto files fail');
```
alexander-fenster: ...why? (`err.toString()`)

typescript/src/start_script.ts
```diff
@@ -0,0 +1,104 @@
...
+    .split('\n');
+  // skip common gax proto files
+  const requiredProtos: string[] = [];
+  protos.forEach(proto => {
+    if (!fs.existsSync(path.join(GOOGLE_GAX_PROTOS_DIR, proto))) {
+      requiredProtos.push(proto);
+    }
+  });
+  requiredProtos.forEach(proto => {
+    protoDirs.forEach(dir => {
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: we could have multiple proto directories (several `-I`), so it might be reasonable to loop the protoDirs

typescript/src/start_script.ts
```diff
@@ -0,0 +1,104 @@
...
+      `${protoFiles} ` +
+      `--typescript_gapic_out=${outputDir} `
+  );
+} catch (err) {
+  console.error('Protoc command fails.');
+  process.exit(1);
+}
+
+// create protos folder to copy proto file
+const COPY_PROTO_DIR = path.join(outputDir, 'protos');
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: got it, thanks.

typescript/src/start_script.ts
xiaozhenliu-gg5: Sure, just console err.toString()

typescript/src/start_script.ts
xiaozhenliu-gg5: Cool suggestion! it definitely makes code more more elegant :)

typescript/src/start_script.ts
```diff
@@ -0,0 +1,104 @@
...
+try {
+  execSync(
+    `protoc ` +
+      `-I${GOOGLE_GAX_PROTOS_DIR} ` +
+      `${protoDirsArg} ` +
+      `${protoFiles} ` +
+      `--typescript_gapic_out=${outputDir} `
+  );
+} catch (err) {
+  console.error('Protoc command fails.');
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: That makes totally sense, thanks!

typescript/src/start_script.ts
xiaozhenliu-gg5: Played around two options a little bit and use `execFileSync` to take parameters with spaces

typescript/src/start_script.ts
```diff
@@ -0,0 +1,96 @@
...
+if (Array.isArray(argv._)) {
+  protoFiles.push(...argv._);
+} else {
+  protoFiles.push(argv._);
+}
+
+// run protoc command to generate client library
+const protocCommand = [
+  `-I${GOOGLE_GAX_PROTOS_DIR}`,
+  `${protoDirsArg}`,
```
alexander-fenster: Wait, this is still incorrect. Just assume there is a space in the path (which is very possible on Windows and sometimes on other systems too).

```
gapic-generator-typescript -I"C:/Documents and Settings/username/protos"
```

so `protoDirsArgs` will become `-IC:/Documents and Settings/username/protos` and, since you convert it to a string in this line, will become separate parameters `-IC:/Documents`, `and`, `Settings/username/protos`.

To fix it, each element of `protoDirsArg` should become an element of `protocCommand` array.

typescript/src/start_script.ts
```diff
@@ -0,0 +1,96 @@
...
+try {
+  execFileSync(`protoc`, protocCommand, { stdio: 'inherit' });
+} catch (err) {
+  console.error(err.toString());
+  process.exit(1);
+}
+
+// create protos folder to copy proto file
+const copyProtoDir = path.join(outputDir, 'protos');
+if (!fs.existsSync(copyProtoDir)) {
```
alexander-fenster: Let's move this `if` under the `try` below since `mkdir` can also fail.

PR #51, opened 2019-10-15
test: expand baseline test 

xiaozhenliu-gg5:
1. Add generated proto list checking
2. Add generated client config checking


typescript/test/unit/baseline.ts
```diff
@@ -46,15 +58,38 @@ const CLIENT_LIBRARY_BASELINE = path.join(
...
+  cwd,
+  'typescript',
+  'test',
+  'testdata',
+  'showcase',
+  'src',
+  'echo_client_config.json.txt'
+);
+
+const CLIENT_PROTO_LIST_BASELINE = path.join(
```
alexander-fenster: OK so this works, now let's make one more logical move and let's just put all the baseline files into the same folder structure and compare all of them.

```
testdata/
  showcase/
    package.json.txt
    src/
      index.ts.txt
        v1/
          echo_client.ts.txt
```

(maybe even rename `.txt` to `.baseline` to make it clear)

Then just compare two folders recursively.

typescript/test/unit/baseline.ts
```diff
@@ -79,10 +78,57 @@ describe('CodeGeneratorTest', () => {
...
-      );
+      const outputFiles = fs.readdirSync(OUTPUT_DIR);
+      // store every item (file or directory with full path in output dir and baseline dir) in stack, pop once a time
+      const protoItemStack: Item[] = [];
+      outputFiles.forEach(file => {
+        const fileFullPath = path.join(OUTPUT_DIR, file);
+        const baselinePath = path.join(BASELINE_DIR, file);
+        protoItemStack.push(new Item(file, fileFullPath, baselinePath));
+      });
+      console.warn(protoItemStack);
```
alexander-fenster: please remove :)

typescript/test/unit/baseline.ts
```diff
@@ -79,10 +78,57 @@ describe('CodeGeneratorTest', () => {
...
+      while (protoItemStack.length !== 0) {
+        const item = protoItemStack.pop();
+        if (!item) continue;
+        // if item is a file, compare it with baseline
+        if (fs.lstatSync(item.outputPath).isFile()) {
+          checkIdenticalFile(
+            item.outputPath,
+            item.baselinePath + BASELINE_EXTENSION
+          );
+        } else {
```
alexander-fenster: let's explicitly check `isDir()` here, there are so many weird types of files

typescript/test/unit/baseline.ts
```diff
@@ -79,10 +78,57 @@ describe('CodeGeneratorTest', () => {
...
+          // if item is a directory, loop the folder and put every item in stack again
+          const items2level = fs.readdirSync(item.outputPath);
+          items2level.forEach(item2level => {
+            const fileFullPath2level = path.join(item.outputPath, item2level);
```
alexander-fenster: style: these three lines are very similar to the three lines above (lines 85-87).
Maybe moving the `path.join` logic to the `Item` constructor can simplify this code a little bit.

typescript/test/unit/baseline.ts
```diff
@@ -79,10 +78,57 @@ describe('CodeGeneratorTest', () => {
...
 });
+
+function checkIdenticalFile(outputFullPath: string, baselineFullPath: string) {
+  if (fs.existsSync(baselineFullPath)) {
+    assert.deepStrictEqual(
+      fs.readFileSync(outputFullPath).toString(),
+      fs.readFileSync(baselineFullPath).toString()
+    );
+  } else {
+    process.exit(1);
```
alexander-fenster: It's too much - if a file does not exist, it's just a test failure. `assert(fs.existsSync(baselineFullPath))` should be enough.

xiaozhenliu-gg5: Cool, sounds good to me :)

typescript/test/unit/baseline.ts
```diff
@@ -79,10 +78,57 @@ describe('CodeGeneratorTest', () => {
...
+      const protoItemStack: Item[] = [];
+      outputFiles.forEach(file => {
+        const fileFullPath = path.join(OUTPUT_DIR, file);
+        const baselinePath = path.join(BASELINE_DIR, file);
+        protoItemStack.push(new Item(file, fileFullPath, baselinePath));
+      });
+      console.warn(protoItemStack);
+      while (protoItemStack.length !== 0) {
+        const item = protoItemStack.pop();
+        if (!item) continue;
```
alexander-fenster: how is this condition possible?

typescript/test/unit/baseline.ts
xiaozhenliu-gg5: Got it, made changes.

typescript/test/unit/baseline.ts
xiaozhenliu-gg5: I think the type checking complains here that `item` might be undefined, so I added this condition to make sure it's passing the check.

alexander-fenster:
One more thing to add here would be to check that all baseline files are actually generated. Now you're iterating over the generated file, so if we stop producing `echo_client.ts` at all because of a bug, this test will not catch it.

package.json
```diff
@@ -53,7 +53,9 @@
     "typescript": "~3.6.0"
   },
   "dependencies": {
+    "@types/fs-extra": "^8.0.1",
```
alexander-fenster: `@types` should go to `devDependencies`. Also, `fs-extra`, if it's only used in tests, should also go to `devDependencies`.

typescript/test/unit/baseline.ts
```diff
@@ -14,18 +14,23 @@
...
-  'echo_client.ts'
+const BASELINE_DIR = path.join(
+  __dirname,
+  '..',
+  '..',
+  '..',
+  'typescript',
+  'test',
+  'testdata',
+  'showcase'
```
alexander-fenster: Let's remove `showcase` from here, then it will be very easy to add more full-sized baseline tests for other libraries.

alexander-fenster: (I mean, the showcase test will start from `path.join(BASELINE_DIR, 'showcase')`, some other baseline test - the multiple protos one - can start from some other folder, etc., all sharing the same comparison code.

typescript/test/unit/baseline.ts
```diff
@@ -79,10 +77,83 @@ describe('CodeGeneratorTest', () => {
...
+        }
+      }
+      assert.strictEqual(fileStack.length, 0);
     });
   });
 });
+
+function checkIdenticalFile(outputFullPath: string, baselineFullPath: string) {
+  assert(fs.existsSync(baselineFullPath))
+  assert.deepStrictEqual(
```
alexander-fenster: This will generate a LOT of output. Also, it will fail on the first assert.
Maybe we can just save all the failures (filenames of non-existent files and filenames that don't match) to an array and then in the test just `assert(checkResultArray.length === 0)`.

Just thinking about the user: the output of `npm test` should be small enough to understand what's going on, and how many baseline files differ.

typescript/test/unit/multi_proto.ts
```diff
@@ -47,13 +47,14 @@ const PROTOLIST_LIBRARY_BASELINE = path.join(
...
   'keymanagementservice_proto_list.json'
 );
 const SRCDIR = path.join(cwd, 'build', 'src');
 const CLI = path.join(SRCDIR, 'cli.js');
 const PLUGIN = path.join(SRCDIR, 'protoc-gen-typescript_gapic');
 
-describe('Proto List Generate Test', () => {
+describe('MultiProtoListGenerateTest', () => {
   describe('Generate Client library', () => {
     it('Generated proto list should have same output with baseline.', function() {
```
alexander-fenster: Can also be converted to using the same baseline logic (especially if you extract it to a separate file).

typescript/test/util.ts
```diff
@@ -0,0 +1,104 @@
+import * as assert from 'assert';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+
+const BASELINE_EXTENSION = '.baseline';
+export function comapreToBaseline(outpurDir: string, baselineDir: string) {
```
alexander-fenster: `compareToBaseline` (typo, it says `comapre`)

typescript/test/unit/multi_proto.ts
xiaozhenliu-gg5: done.

typescript/test/unit/baseline.ts
xiaozhenliu-gg5: Sure!

package.json
xiaozhenliu-gg5: done.

typescript/test/unit/baseline.ts
xiaozhenliu-gg5: Very helpful suggestions! Output error of not existing files and not identical files separately, which will not block the test.

typescript/test/util.ts
xiaozhenliu-gg5: done.

xiaozhenliu-gg5:
Thanks Alex for the super valuable comments!

typescript/test/util.ts
```diff
@@ -0,0 +1,103 @@
...
+  let fileStack: string[] = [];
+  const dirStack: string[] = [];
+  putAllBaselineFiles(baselineDir, fileStack, dirStack);
+  // store every item (file or directory with full path in output dir and baseline dir) in stack, pop once a time
+  const protoItemStack: Item[] = [];
+  putItemToStack(protoItemStack, outpurDir, baselineDir);
+  while (protoItemStack.length !== 0) {
+    const item = protoItemStack[0];
+    protoItemStack.shift();
+    // if item is a file, compare it with baseline
```
xiaozhenliu-gg5: seems array.pop() is removing element from the end, debugged a little bit to find it. So use shift here to make sure we remove the first element every time.

alexander-fenster: Then it's a queue and not a stack

alexander-fenster:
@xiaozhenliu-gg5 I merged the code and updated your branch to make tests pass, please do `git pull` before making changes.

typescript/test/util.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (fs.lstatSync(item.outputPath).isFile()) {
+      const identical = checkIdenticalFile(
+        item.outputPath,
+        item.baselinePath + BASELINE_EXTENSION
+      );
+      // if two files are identical or it's generated properly, filter it from the stack.
+      if (identical !== 2) {
```
alexander-fenster: 2 is a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)), please use meaningful constants.

typescript/test/util.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+  assert.strictEqual(fileStack.length, 0);
+}
+
+function checkIdenticalFile(
+  outputFullPath: string,
+  baselineFullPath: string
+): number {
+  if (!fs.existsSync(baselineFullPath)) {
+    console.error(baselineFullPath + 'is not generated. ');
```
alexander-fenster: This case is not covered by an assertion.

alexander-fenster: At least it seems so - please check the CircleCI logs of the recent run, the start script test should've failed because of the missing file but it passes. I'm not sure why the proto file is missing btw.

typescript/test/unit/multi_proto.ts
```diff
@@ -17,16 +17,11 @@ import { execSync } from 'child_process';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as rimraf from 'rimraf';
+import { compareToBaseline } from '../util';
 
 const cwd = process.cwd();
 
 const OUTPUT_DIR = path.join(cwd, '.multi-proto-test-out');
```
alexander-fenster: Maybe make it `.test-out-keymanager` and `.test-out-showcase` to make it easier to find?

alexander-fenster:
I think we are almost there!

typescript/test/util.ts
xiaozhenliu-gg5: yes, then I probably should get item from the end.

typescript/test/unit/multi_proto.ts
xiaozhenliu-gg5: Sure!

typescript/test/util.ts
alexander-fenster: It does not really matter if you take it from the beginning or from the end (it only affects the order of files but at the end produces the same results), but the variable name should match the way you're working with it.

typescript/test/util.ts
xiaozhenliu-gg5: yes, you are right, it should check whether `baselineFullPath ` exists or not. And after adding assertion to that, circleCI will fail coz it could not find `../../echo.proto.baseline` in baseline directory.

typescript/test/util.ts
xiaozhenliu-gg5: Right! Thanks! was wondering how should I deal with that.

typescript/test/util.ts
xiaozhenliu-gg5: And one more logic added, if it's a `.proto` file, we just skip it without comparing because baseline does not have protos, which makes tests all green now.

typescript/test/util.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+  assert.strictEqual(fileStack.length, 0);
+}
+
+function checkIdenticalFile(
+  outputFullPath: string,
+  baselineFullPath: string
+): number {
+  if (!outputFullPath.endsWith('.proto')) {
+    assert(fs.existsSync(baselineFullPath));
```
alexander-fenster: Let's not assert this, because it stops other files from being checked. Let's just return `IDENTICAL_FILE` to skip the check for `.proto` files.

xiaozhenliu-gg5: Sounds good, take proto file as identical and log error if the file does bot exist. Return boolean for the util method `equalToBaseline` and only assert there. Thanks!

PR #52, opened 2019-10-16
refactor: make sure templates can be located anywhere

alexander-fenster:
We use relative paths to templates which makes it hard to use the generator when the current working directory is different. Fixing that by passing the base path parameter to `nunjucks` in `nunjucks.configure`.

Not including any tests here because this is a refactor and existing tests cover this code (and they still work).

(also, ran `gts fix` that changed one unrelated function)

### DIFF
--- a/typescript/src/generator.ts
+++ b/typescript/src/generator.ts
@@ -21,7 +21,13 @@ import { API } from './schema/api';
 import { processTemplates } from './templater';
 import { commonPrefix } from './util';
 
-const templateDirectory = 'templates/typescript_gapic';
+const templateDirectory = path.join(
+  __dirname,
+  '..',
+  '..',
+  'templates',
+  'typescript_gapic'
+);
 // If needed, we can make it possible to load templates from different locations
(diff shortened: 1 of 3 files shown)


typescript/src/generator.ts
```diff
@@ -83,8 +89,6 @@ export class Generator {
     const api = this.buildAPIObject();
     await this.processTemplates(api);
```
xiaozhenliu-gg5: the `processTemplate` function take basePath, we only give api here, does this effect anything? I'm wondering how we give the path to template since we don't have default `basePath` value in `processTemplate`. Thanks!

alexander-fenster: It's my bad naming. This line calls `this.processTemplates()` (a method in the class), but we also have `import { processTemplates } from './templater'` which is a different function (that takes 2 arguments).

xiaozhenliu-gg5: Oh, my bad, I just looked into the changing part, not the whole file, thanks for the clarification!

Issue #58, opened 2019-10-16
should be possible to create a client without options


alexander-fenster:
Right now it requires to pass `opts`. Need to change some generated code inside to allow `opts?`.

PR #65, opened 2019-10-21
feat: allow constructors with no parameters

alexander-fenster:
Some changes allowing to call constructors with no parameters in a type safe way.


alexander-fenster:
Fixes #58

PR #66, opened 2019-10-22
chore: add comments on service & method level

### LINKED ISSUE #17: comments
We should provide comments for generated methods. The code samples are pretty straightforward, but the first line (method description) should be taken from input proto.

Example: https://github.com/googleapis/gax-nodejs/blob/7c8000b8ac9d48f27711912d5acc2382d5839826/test/fixtures/google-gax-packaging-test-app/src/v1beta1/echo_client.js#L260-L299

xiaozhenliu-gg5:
fix #17

### DIFF
--- a/typescript/src/schema/comments.ts
+++ b/typescript/src/schema/comments.ts
@@ -48,8 +48,8 @@ export class CommentsMap {
   getCommentsMap() {
     return this.comments;
   }
-  getServiceComment(service: plugin.google.protobuf.IServiceDescriptorProto) {
-    return this.comments[service.name! + 'Service'].trim();
+  getServiceComment(serviceName: string) {
+    return this.comments[serviceName + 'Service'].trim();
   }
   getMethodComments(methodName: string): string {
     return this.comments[methodName].trim();
--- a/typescript/src/schema/comments.ts
+++ b/typescript/src/schema/comments.ts
@@ -1,0 +1,51 @@
+import * as plugin from '../../../pbjs-genfiles/plugin';
+export interface Comments {
+  [name: string]: string;
+}
+
+export class CommentsMap {
+  comments: Comments = {};
+
+  constructor(fd: plugin.google.protobuf.IFileDescriptorProto) {
+    const commentsMap: Comments = {};
+    if (fd && fd.sourceCodeInfo && fd.sourceCodeInfo.location) {
+      const locations = fd.sourceCodeInfo.location;
+      locations.forEach(location => {
+        if (location.leadingComments !== null) {
+          // p is an array with format [f1, i1, f2, i2, ...]
+          // - f1 refers to the protobuf field tag
+          // - if field refer to by f1 is a slice, i1 refers to an element in that slice
+          // - f2 and i2 works recursively.
+          // So, [6, x] refers to the xth service defined in the file,
+          // since the field tag of Service is 6.
+          // [6, x, 2, y] refers to the yth method in that service,
+          // since the field tag of Method is 2.
+          const p = location.path!;
+          if (p.length === 2 && p[0] === 6) {
+            if (fd.service && fd.service[p[1]] && fd.service[p[1]].name) {
+              commentsMap[fd.service[p[1]].name! + 'Service'] =
+                location.leadingComments || '';
+            }
+          } else if (p.length === 4 && p[2] === 2) {
+            if (
+              fd.service &&
+              fd.service[p[1]] &&
+              fd.service[p[1]].method &&
+              fd.service[p[1]].method![p[3]] &&
+              fd.service[p[1]].method![p[3]].name
+            ) {
+              const name = fd.service[p[1]].method![p[3]].name!;
+              if (!commentsMap[name]) {
+                commentsMap[name] = location.leadingComments || '';
+              }
+            }
+          }
+        }
+      });
+      this.comments = commentsMap;
+    }
+  }
+  getCommentsMap() {
+    return this.comments;
+  }
+}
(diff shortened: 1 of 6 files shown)


typescript/src/schema/proto.ts
```diff
@@ -237,12 +258,54 @@ function augmentService(
...
+            location.leadingComments || '';
+        }
+      } else if (p.length === 4 && p[2] === 2) {
+        if (
+          fd.service &&
+          fd.service[p[1]] &&
+          fd.service[p[1]].method &&
+          fd.service[p[1]].method![p[3]] &&
+          fd.service[p[1]].method![p[3]].name
+        ) {
```
xiaozhenliu-gg5: Hi Alex, is there a better way to check `fd.service[p[1]].method![p[3]].name` exists instead of checking every layer?  Even if I check layer by layer, ts still complains about `fd.service[p[1]].method` might be `null` or `undefined` so I put `fd.service[p[1]].method!` below to make it pass. But really wondering why it fails type checking. Thanks!

alexander-fenster: It will appear soon with [optional chaining](https://github.com/microsoft/TypeScript/issues/30167) but not right now. Let's leave it as here for now, and we'll be able to refactor this when this TypeScript feature lands.

Also, the type checks are static, they cannot apply to the random array element like `p[1]`. If you saved each individual part to a variable, it would've worked (e.g. `const method = fd.service[p[1]].method; if (method && method[p[3]]) { const whatever = method[p[3]]; if (whatever) { const name = whatever.name; }` but that does not make much sense to me.

Other thing is... can we change `1` and `3` to constants if it makes sense?

typescript/src/schema/proto.ts
```diff
@@ -162,7 +167,18 @@ function toLRInterface(type: string, inputType: string) {
   return inputType.replace(/\.([^.]+)$/, '.I' + type);
 }
 
-function augmentMethod(messages: MessagesMap, method: MethodDescriptorProto) {
+function getComments(
+  method: MethodDescriptorProto,
+  comments: CommentsMap
+): string {
+  return comments[method.name!].trim();
```
alexander-fenster: minor suggestion: let's make `CommentsMap` a class and this `getComments` function will become its method. `getCommentsMap` will then become a constructor. It can go to a separate file `comments.ts`.

typescript/src/schema/proto.ts
alexander-fenster: After reading the code, maybe we don't really change numbers to constants here. Just as I wrote in another comment, let's move all this stuff to a separate `comments.ts` and leave it there as it is.

xiaozhenliu-gg5: Oh I did not know we have this new `optional chaining `feature coming, sounds awesome! 
and sure thing, I will move comment related stuff to a separate ts which makes more sense and also make the code cleaner.

typescript/src/schema/proto.ts
xiaozhenliu-gg5: Will do!

typescript/src/schema/proto.ts
```diff
@@ -153,6 +156,18 @@ function pagingResponseType(
 function toInterface(type: string) {
   return type.replace(/\.([^.]+)$/, '.I$1');
 }
+function getMethodComments(
```
alexander-fenster: OK so the next step of making comments a black box is to make this function a public method of `Comments` class. You don't need to expose the comments map, just keep an instance of `comments`, and do `comments.getMethodComments(methodName)` when needed.

alexander-fenster:
Just a general note - the idea of making something a separate class is to hide its implementation from its users behind its interface. Right now the comments are implemented as an object (a map) and `getMethodComments` knows about it, which exposes the comments implementation to the users of that class. By making `getMethodComments` a public method of `Comments` class we're limiting the knowledge of implementation just to `comments.ts`, making it a black box for this class' users.

typescript/src/schema/proto.ts
xiaozhenliu-gg5: It totally makes sense to me. I tried to put everything in `comments` class: `getMethodComments` `getServiceComments` and `getCommentsMap`, but I realized that the argument `method: MethodDescriptorProto` requires type from `proto.ts` and `proto.ts` wants to use `comments.ts` which is a circle. 

Or I can pass `method.name` to `getMethodComments`.

typescript/src/schema/comments.ts
```diff
@@ -0,0 +1,57 @@
...
+      this.comments = commentsMap;
+    }
+  }
+  getCommentsMap() {
+    return this.comments;
+  }
+  getServiceComment(service: plugin.google.protobuf.IServiceDescriptorProto) {
+    return this.comments[service.name! + 'Service'].trim();
+  }
+  getMethodComments(methodName: string): string {
```
alexander-fenster: nit: to make it logical, if `getServiceComment` accepts a descriptor proto, can we make `getMethodComments` accept a descriptor proto as well?

xiaozhenliu-gg5: like last comment, `MethodDescriptorProto` is defined in `proto.ts`, if we require `MethodDescriptorProto` in `comments.ts` and `proto.ts` require `comments.ts`, that would be in circle. 
`service` is different because `service: plugin.google.protobuf.IServiceDescriptorProto` is from `plugin`, while `method: MethodDescriptorProto`.

I agree to put service as an argument, but not sure there is a good way to do that. Thanks! Would appreciate if you can give some advice!

PR #67, opened 2019-10-23
chore: change generated filename to snake_case

### LINKED ISSUE #55: should use snake_case for filenames
In the templater, we are now using `toLowerCase` for generating filenames (e.g. `texttospeech_client.ts`). For compatibility with gapic-generator, we need to use snake_case (`text_to_speech_client.ts`). We have a function to generate snake_case in `util.ts` so it should be a quick change.

xiaozhenliu-gg5:
fix #55


typescript/src/templater.ts
```diff
@@ -68,7 +69,7 @@ function processOneTemplate(
     for (const service of api.services) {
       result.push(
         renderFile(
-          outputFilename.replace(/\$service/, service.name!.toLowerCase()),
+          outputFilename.replace(/\$service/, service.name!.toSnakeCase()),
```
alexander-fenster: There must be a change in the baseline, the place where we import the renamed JSON file.

xiaozhenliu-gg5: got it, thanks!

PR #71, opened 2019-10-24
feat: respect resource annotation

### LINKED ISSUE #59: PathTemplate support
```
    // This API contains "path templates"; forward-slash-separated
    // identifiers to uniquely identify resources within the API.
    // Create useful helper objects for these.
    this._pathTemplates = {
      locationPathTemplate: new gaxModule.PathTemplate(
        'projects/{project}/locations/{location}'
      ),
    };
```

```
  /**
   * Return a fully-qualified location resource name string.
   *
   * @param {String} project
   * @param {String} location
   * @returns {String}
   */
  locationPath(project, location) {
    return this._pathTemplates.locationPathTemplate.render({
      project: project,
      location: location,
    });
  }
```

```
  /**
   * Parse the glossaryName from a glossary resource.
   *
   * @param {String} glossaryName
   *   A fully-qualified path representing a glossary resources.
   * @returns {String} - A string representing the location.
   */
  matchLocationFromGlossaryName(glossaryName) {
    return this._pathTemplates.glossaryPathTemplate.match(glossaryName)
      .location;
  }
```

xiaozhenliu-gg5:
fix #59


templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -480,4 +493,23 @@ export class {{ service.name }}Client {
     return this._innerApiCalls.{{ method.name.toCamelCase() }}(request, options, callback);
   } 
 {%- endfor %}
-}
+{%- if (service.pathTemplate.length > 0) %}
+  // --------------------
+  // -- Path templates --
+  // --------------------
+{%- for template in service.pathTemplate %}
+  {{ template.type.toLowerCase() }}Path({{ template.params.addStringType() }}){
```
alexander-fenster: I suspect it might be `toCamelCase()` here - do we have any resources with multiple words?

typescript/src/schema/proto.ts
```diff
@@ -436,6 +442,35 @@ function augmentService(
       '.google.api.oauthScopes'
     ].split(',');
   }
+  augmentedService.pathTemplate = [];
+  for (const property in messages) {
+    if (messages.hasOwnProperty(property)) {
```
alexander-fenster: It's easier to do `for (const property of Object.keys(messages))` without any additional filtering.

typescript/src/schema/proto.ts
```diff
@@ -436,6 +442,35 @@ function augmentService(
...
+      const m = messages[property];
+      if (m && m.options) {
+        const option = m.options;
+        if (option && option['.google.api.resource']) {
+          const opt = option['.google.api.resource'];
+          const onePathTemplate = opt as ResourceDescriptor;
```
alexander-fenster: Maybe merge these two lines together? `const resourceDescriptor = option['.google.api.resource'] as ResourceDescriptor;` - there is no need to introduce a new variable.

typescript/src/util.ts
```diff
@@ -63,6 +63,12 @@ String.prototype.capitalize = function(this: string): string {
   }
   return this[0].toUpperCase() + this.slice(1);
 };
+String.prototype.capitalizeFirst = function(this: string): string {
```
alexander-fenster: We already have `.capitalize` that does _exactly_ this :)

templates/typescript_gapic/src/$version/$service_client.ts.njk
xiaozhenliu-gg5: I'm not sure we have multiple-words resources, but yea I think camel case is better here.

typescript/src/schema/proto.ts
xiaozhenliu-gg5: got it, will make changes

typescript/src/schema/proto.ts
xiaozhenliu-gg5: we use this `opt` below to check its type and pattern, not sure we want to combine them into one line, so that would be `option['.google.api.resource']` everywhere

typescript/src/util.ts
xiaozhenliu-gg5: On nice, thought it's capitalizing all letters, great! I will remove this helper function, thanks!

typescript/test/testdata/texttospeech/src/v1/text_to_speech_client.ts.baseline
```diff
@@ -313,4 +313,4 @@ export class TextToSpeechClient {
     return this._innerApiCalls.synthesizeSpeech(request, options, callback);
   }
 
-}
```
alexander-fenster: Please add new lines to the template and baseline files.
image

typescript/src/schema/proto.ts
```diff
@@ -436,6 +442,35 @@ function augmentService(
...
     ].split(',');
   }
+  augmentedService.pathTemplate = [];
+  for (const property of Object.keys(messages)) {
+    const m = messages[property];
+    if (m && m.options) {
+      const option = m.options;
+      if (option && option['.google.api.resource']) {
+        const opt = option['.google.api.resource'];
+        const onePathTemplate = option[
```
alexander-fenster: OK, let's do `opt as ResourceDescriptor` then :)  It raises questions when you access `options['.google.api.resource'] twice in the two consecutive lines.

typescript/src/util.ts
```diff
@@ -105,3 +105,10 @@ String.prototype.toSnakeCase = function(this: string): string {
   }
   return words.join('_');
 };
+
+Array.prototype.addStringType = function(this: string[]): string[] {
+  if (this.length === 0) {
+    return this;
+  }
+  return this.map(p => p + ': string');
```
alexander-fenster: Here we basically incorporate a part of TypeScript output into the source code (and not the templates). We are trying not to do it that often. (the generator code actually has nothing to do with TypeScript output, and it's theoretically possible to generate any other language by changing just the templates)

Can we do template iteration with a `joiner()` instead?

xiaozhenliu-gg5: Sure thing, great suggestion!

alexander-fenster:
The fact that we have no baseline changes means that our tests do not cover this at all :)  Let's add a real API (e.g. `translate`) into our baseline tests within the next PR (the one that brings in common resources).

PR #73, opened 2019-10-24
feat: generate JSON service config based on gRPC service config

alexander-fenster:
The following huge (I'm sorry!) PR implements the following changes:

- make it possible to pass gRPC service config to the generator using `--grpc-service-config` option
- generate the service config JSON in GAPIC format based on the gRPC service config passed
- added tests for this functionality

Also, some smaller changes:
- changed a way we run the generator from the start script (a little refactor caused by all these command line options stuff)
- added `--help` to the start script
- fixed random permissions issue in CircleCI
- random small fixes here and there


typescript/src/schema/proto.ts
```diff
@@ -1,9 +1,30 @@
...
+  plugin.google.rpc.Code.DEADLINE_EXCEEDED,
+  plugin.google.rpc.Code.UNAVAILABLE,
+];
+const defaultParametersName = 'default';
+const defaultParameters = {
+  initial_retry_delay_millis: 100,
+  retry_delay_multiplier: 1.3,
+  max_retry_delay_millis: 60000,
+  // note: the following four parameters are unused but currently required by google-gax.
+  // setting them to some big safe default values.
```
noahdietz: Why are they required by google-gax? and is this something that is standard across languages (if you know)?

This is more for my own edification. I assumed supplying a 0 value for these would be OK.

typescript/src/schema/proto.ts
```diff
@@ -17,10 +38,114 @@ interface MethodDescriptorProto
...
+    retryableStatusCodes: plugin.google.rpc.Code[]
+  ): string {
+    // generate an unique readable name for the given retryable set of codes
+    const sortedCodes = retryableStatusCodes.sort(
+      (a, b) => Number(a) - Number(b)
+    );
+    const uniqueName = sortedCodes
+      .map(code => this.codeEnumMapping[code])
+      .join('_')
+      .toSnakeCase();
```
noahdietz: I might be misunderstanding this code, but it looks like it is concatenating the retry codes with a `_` to use as the name. I'm not sure that this name would actually be unique. `INTERNAL_UNAVAILABLE_UNKNOWN` as a code set name might not be unique.

Am I misunderstanding it?

typescript/src/schema/proto.ts
```diff
@@ -17,10 +38,114 @@ interface MethodDescriptorProto
...
+    return uniqueName;
+  }
+
+  private buildUniqueParamsName(params: {}): string {
+    // generate an unique not so readable name for the given set of parameters
+    return objectHash(params);
```
noahdietz: I'm not super familiar with hashing, but if I have two retry policies that contain all of the same values, but for different MethodConfigs (maybe there is something else different about the MethodConfig other than the RetryPolicy), how do we know that this will in fact be unique?

typescript/test/testdata/keymanager/src/v1/key_management_service_client_config.json.baseline
```diff
@@ -16,124 +14,101 @@
...
           "initial_rpc_timeout_millis": 20000,
-          "rpc_timeout_multiplier": 1.0,
+          "rpc_timeout_multiplier": 1,
           "max_rpc_timeout_millis": 20000,
           "total_timeout_millis": 600000
         }
       },
       "methods": {
         "ListKeyRings": {
-          "timeout_millis": 60000,
```
noahdietz: It seems like all of the methods marked non_idempotent are losing their timeout. I'm unsure about this specific test if that is intended or not.

typescript/src/schema/proto.ts
alexander-fenster: They are technically not required by google-gax logic, but are listed as required in the TypeScript type [here](https://github.com/googleapis/gax-nodejs/blob/master/src/gax.ts#L578). It will require a gax release to remove them from there, and when it's done, I will remove them from the generator as well.

typescript/src/schema/proto.ts
alexander-fenster: You are understanding it right. The idea is that I need some human readable name for an array of codes, e.g. `[UNAVAILABLE, INTERNAL, UNKNOWN]`, so I first sort them, then join with `_`.

It's indeed possible that the resulting code won't be unique if the `google.rpc.Code` enum has elements with `_`, e.g. (imaginary example) `UNKNOWN_ERROR`, `UNKNOWN`, and `ERROR`. So I looked at the actual contents of the enum and found that–if I'm not mistaken–the uniqueness of the code built by this method still holds:

```
{
  OK: 0,
  CANCELLED: 1,
  UNKNOWN: 2,
  INVALID_ARGUMENT: 3,
  DEADLINE_EXCEEDED: 4,
  NOT_FOUND: 5,
  ALREADY_EXISTS: 6,
  PERMISSION_DENIED: 7,
  UNAUTHENTICATED: 16,
  RESOURCE_EXHAUSTED: 8,
  FAILED_PRECONDITION: 9,
  ABORTED: 10,
  OUT_OF_RANGE: 11,
  UNIMPLEMENTED: 12,
  INTERNAL: 13,
  UNAVAILABLE: 14,
  DATA_LOSS: 15
}
```

It's technically possible that new codes will be added that will make this method generate non-unique codes but the possibility of it is very low in my opinion (and concatenating with `_` makes the keys readable).

typescript/test/testdata/keymanager/src/v1/key_management_service_client_config.json.baseline
alexander-fenster: That's because these two baseline tests (showcase and KMS) don't have gRPC service config. Previously, `timeout_millis` was hardcoded here.

The `texttospeech` baseline has gRPC service config and `timeout_millis` is set correctly.

typescript/src/schema/proto.ts
alexander-fenster: Here I only care about the contents of `RetryPolicy` so it should be fine.

Actually, I think the logic of this class is complicated enough to be covered by unit tests, not just by baseline tests. I will add some unit tests that will likely explain the use cases better than just the code :)

typescript/test/testdata/keymanager/src/v1/key_management_service_client_config.json.baseline
noahdietz: Ah ok thank you for the clarification.

typescript/src/schema/proto.ts
noahdietz: Ok so if two different MethodConfigs have the same set of retryable codes `[UNAVAILABLE, UNKNOWN]`, their RPCs would both use the same generated retry code set named `UNAVAILABLE_UNKNOWN` and that's OK because the set of the retryable codes between the two MethodConfigs are the same. Gotchya.

typescript/src/schema/proto.ts
noahdietz: SGTM, thanks

typescript/src/schema/proto.ts
noahdietz: Oh that makes sense, thanks.

PR #76, opened 2019-10-25
fix: use require.resolve to locate google-gax

alexander-fenster:
It basically can be anywhere if we use `npm install -g` to install `gapic-generator-typescript`. `require.resolve` is the right way of figuring out the location of the dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
   },
   "files": [
     "build/src",
-    "pbjs-genfiles"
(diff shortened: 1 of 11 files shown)


alexander-fenster:
@bcoe Is `require.resolve` the right way of locating the dependency? I figured if we install it with `npm i -g` it's not always in our own `node_modules`.

PR #77, opened 2019-10-27
feat: docker interface

### LINKED ISSUE #19: docker interface
TypeScript micro-generator should be shipped as a Docker image. It should auto-release from CircleCI.

alexander-fenster:
Fixes #19.

Introducing Docker interface for `gapic-generator-typescript`. Compatible with Python's [gapic.sh](https://github.com/googleapis/gapic-generator-python/blob/master/gapic.sh). Follows [AIP-4290](https://aip.dev/client-libraries/4290).

Note 1: to make it possible for `--grpc-service-config` to pass JSON config file relative to `--in`, the script inside Docker does `cd /in` before executing `gapic-generator-typescript`.

Note 2: since we need to perform post processing (copy proto files into the result directory), `gapic-generator-typescript` is actually a small wrapper calling `protoc`, and the actual generator is, in turn, a `protoc` plugin.


docker/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN apk update && apk add libc6-compat && rm -f /var/cache/apk/*
+ENV LD_LIBRARY_PATH /lib64:$LD_LIBRARY_PATH
+RUN cd /root/files && \
+    wget https://github.com/protocolbuffers/protobuf/releases/download/v3.10.0/protoc-3.10.0-linux-x86_64.zip
+RUN cd /usr/local && unzip /root/files/protoc-3.10.0-linux-x86_64.zip
+
+# Download a copy of API common protos
+RUN cd /root/files && \
+    wget https://github.com/googleapis/api-common-protos/archive/master.zip
+RUN cd /protos && unzip /root/files/master.zip && chmod -R a+rx /protos
```
noahdietz: These can be extracted from the api-common-protos docker image. However I'm not sure the last time they were updated, protoc version or the common protos.

alexander-fenster: Yeah, my concern is that master version on GitHub is newer than the latest release, and I'm not sure about Docker image updates either, so I decided to take it from GitHub master for now, until the dust settles.

noahdietz: OK that's reasonable to me.

noahdietz:
LGTM will let @xiaozhenliu-gg5 approve

PR #80, opened 2019-10-28
fix: make sure npm pack works

alexander-fenster:
Preventing issues like https://github.com/googleapis/nodejs-translate/pull/361.


templates/typescript_gapic/package.json.njk
```diff
@@ -5,6 +5,11 @@
   "author": "Google LLC",
   "description": "{{ api.naming.productName }} client for Node.js",
   "main": "build/src/index.js",
+  "files": [
+    "build/src",
+    "build/protos",
+    "!build/src/**/*.map"
```
bcoe: Can we re-litigate this :smile: I think having the maps is good, especially now that I've added source-map-support to Node.js.

PR #81, opened 2019-10-28
feat: respect common resources

xiaozhenliu-gg5:
1. construct `resource map` at Api level
2. `service.pathTemplates` has all in-use resources including common resources


typescript/src/schema/api.ts
```diff
@@ -83,3 +91,48 @@ export class API {
...
   }
 }
+
+function getResourceMap(
+  fd: plugin.google.protobuf.IFileDescriptorProto[]
+): ResourceMap {
+  const resourceMap: ResourceMap = {};
+  for (const fileDescriptor of fd) {
+    const messages = fileDescriptor
+      .messageType!.filter(message => message.name)
```
alexander-fenster: A few comments related to this code:
1. Naming is a little bit confusing. Let's call the function parameter `fileDescriptors` (in plural), then we can use singular `fd` or `fileDescriptor` anywhere.
2. `messageType` can be `undefined` or `null` (if there are no messages), so let's also filter out only those fds with existing messages.

typescript/src/schema/api.ts
```diff
@@ -83,3 +91,48 @@ export class API {
...
+        const option = m.options;
+        if (option && option['.google.api.resource']) {
+          const opt = option['.google.api.resource'];
+          const oneResource = option[
+            '.google.api.resource'
+          ] as ResourceDescriptor;
+          if (opt.type) {
+            const arr = opt.type.match(/\/([^.]+)$/);
+            if (arr) {
+              oneResource.name = arr[arr.length - 1];
```
alexander-fenster: Since you only have one match group in your regex, it should always be `arr[1]`.

typescript/src/schema/api.ts
```diff
@@ -83,3 +91,48 @@ export class API {
...
+          }
+          const pattern = opt.pattern;
+          if (pattern && pattern[0]) {
+            const params = pattern[0].match(/{[a-zA-Z]+}/g) || [];
+            for (let i = 0; i < params.length; i++) {
+              params[i] = params[i].replace('{', '').replace('}', '');
+            }
+            oneResource.params = params;
+          }
+          resourceMap[opt.type!] = oneResource;
```
alexander-fenster: What if `oneResource` still has name or params undefined? We probably should warn and should not save it.

typescript/src/schema/proto.ts
```diff
@@ -442,32 +448,28 @@ function augmentService(
...
+      const fields = m.field;
+      for (const fieldDescriptor of fields) {
+        if (fieldDescriptor && fieldDescriptor.options) {
+          const option = fieldDescriptor.options;
+          if (option && option['.google.api.resourceReference']) {
+            const resourceReference = option['.google.api.resourceReference'];
+            const type = resourceReference.type;
+            if (!type || !resourceMap[type.toString()]) {
+              console.warn(
+                'resource does not have a type or resource is not in the map'
```
alexander-fenster: 1. Please capitalize the first letter :)

2. This warning does not give any actionable information: what file is it, which resource is bad, what map is it (the user has no idea about our map). The warning messages should be actionable. E.g. `In FILE.proto, message MESSAGE refers to an unknown resource RESOURCE`, or something like this.

typescript/src/schema/api.ts
xiaozhenliu-gg5: Got it!

typescript/src/schema/api.ts
xiaozhenliu-gg5: yes, that makes sense, there is only one element in the array `arr[0]`.

typescript/src/schema/proto.ts
xiaozhenliu-gg5: Sure! Added useful warning messages

typescript/src/schema/api.ts
xiaozhenliu-gg5: very helpful! Added logics to check name and params

PR #83, opened 2019-10-29
feat: support google.api.http annotation

### LINKED ISSUE #74: x-goog-request-params
We don't generate this:

```js
    options.otherArgs = options.otherArgs || {};
    options.otherArgs.headers = options.otherArgs.headers || {};
    options.otherArgs.headers[
      'x-goog-request-params'
      parent: request.parent,
    });
```

It should be pretty straightforward, it's described in https://aip.dev/client-libraries/4222.

xiaozhenliu-gg5:
Found URL-based variables declared `google.api.http` annotation and transcribe these into the x-goog-request-params header in unary calls

fix #74


xiaozhenliu-gg5:
Questions: when generating `dlp` api against the change, it could not find `storage.proto`, for keymanager & translate, there is type incompitibility.

alexander-fenster:
It's good that we run unit tests for the generated libraries :) (please check the KMS CI)

typescript/src/schema/proto.ts
```diff
@@ -288,6 +289,23 @@ function toLRInterface(type: string, inputType: string) {
   return inputType.replace(/\.([^.]+)$/, '.I' + type);
 }
 
+function getHeaderParms(rule: plugin.google.api.IHttpRule): string {
+  const message = rule.post
```
alexander-fenster: Is this the same as `rule.post || rule.delete || rule.get || rule.put || rule.patch`?

typescript/src/schema/proto.ts
```diff
@@ -288,6 +289,23 @@ function toLRInterface(type: string, inputType: string) {
   return inputType.replace(/\.([^.]+)$/, '.I' + type);
 }
 
+function getHeaderParms(rule: plugin.google.api.IHttpRule): string {
```
alexander-fenster: `Params`?

typescript/src/schema/proto.ts
```diff
@@ -288,6 +289,23 @@ function toLRInterface(type: string, inputType: string) {
...
+    ? rule.post
+    : rule.delete
+    ? rule.delete
+    : rule.get
+    ? rule.get
+    : rule.put
+    ? rule.put
+    : rule.patch;
+  if (message){ 
+    const res = message.match('(?<={)(.*)(?==)');
```
alexander-fenster: Positive look-behinds are very rarely used. If you just need to grab something in the curly brackets, just do `/{(.*?)=/`.

typescript/src/util.ts
```diff
@@ -105,3 +105,16 @@ String.prototype.toSnakeCase = function(this: string): string {
   }
   return words.join('_');
 };
+
+String.prototype.CamelCaseBeforeDot = function(this: string): string{
```
alexander-fenster: should start with a lower case `c`

typescript/src/util.ts
```diff
@@ -105,3 +105,16 @@ String.prototype.toSnakeCase = function(this: string): string {
...
+  const words = this.split('.');
+  const res:string[] = [];
+  for(var i = 0; i < words.length - 1; i++){
```
alexander-fenster: much easier: `const res = words.slice(0, -1).map(w => w.toCamelCase())`

typescript/src/util.ts
```diff
@@ -105,3 +105,16 @@ String.prototype.toSnakeCase = function(this: string): string {
...
+    res.push(words[i].toCamelCase());
+  }
+  if(res.length > 0){
+    res.push(words[words.length - 1]);
+    return res.join('!.');
```
alexander-fenster: This is TypeScript specifics in the code. We try not to do that.

Actually, I'm not even sure we need to have this function at all. Can we do it in the templates using a joiner? Dots are also language specific (we would've used `->` in C++ for example).

alexander-fenster:
So, we hit another issue here :(

image

This means that we'll need to improve the way we're generating request object for the unit tests.

typescript/src/schema/proto.ts
alexander-fenster: Also, this function deserves a unit test in `test/unit/proto.ts`.

xiaozhenliu-gg5: yes, this is for getting parameters in the httprule, like `crypto_key.name` would be `[crypto_key, name]`, sure, adding unit test makes perfect sense.

typescript/src/schema/proto.ts
xiaozhenliu-gg5: got it, `/{(.*?)=/` seems to include `=`, use `/{(.*?)(?==)/` instead.

typescript/src/util.ts
xiaozhenliu-gg5: Nice

typescript/src/schema/proto.ts
xiaozhenliu-gg5: it works, thanks! was making it too complicated

typescript/src/util.ts
xiaozhenliu-gg5: made changes, thanks!

typescript/src/util.ts
xiaozhenliu-gg5: this is for typescript type checking, that `request.crypto_key` always exists, so that `request.cropto_key.name` can be called.

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -352,12 +352,24 @@ export class {{ service.name }}Client {
...
+      options = optionsOrCallback as gax.CallOptions;
+    }
     options = options || {};
+{%- if method.headerRequestParams.length > 0 %}
+    options.otherArgs = options.otherArgs || {};
+    options.otherArgs.headers = options.otherArgs.headers || {};
+    options.otherArgs.headers[
+      'x-goog-request-params'
+    ] = gax.routingHeader.fromParams({
+      '{{ method.headerRequestParams.toString().toCamelCase() }}': request.{{ method.headerRequestParams.camelCaseBeforeDot() }} || '',
```
alexander-fenster: It should be snake_case for the left part and camelCase for the right part. Left part is basically taken from proto as is (since it gets passed to gRPC server as a header), right part is converted to camelCase because it's how `protobufjs` processes keys with underscore.

Example here: https://github.com/googleapis/nodejs-logging/blob/dcb48172656de8b449a2c05e8cbe999ebaac90bf/src/v2/metrics_service_v2_client.js#L460

alexander-fenster: Actually, if we discuss it, it would be nice to have it covered by a baseline test :)

templates/typescript_gapic/test/gapic-$service-$version.ts.njk
```diff
@@ -200,7 +226,15 @@ describe('{{ service.name }}Client', () => {
                 projectId: 'bogus', 
             });
             // Mock request
-            const request = {};
+            const request: protosTypes{{ method.inputInterface }} = {};
```
alexander-fenster: I'm thinking that if this same sequence is repeated four times it might make sense to make it a macro (accepting `method` as a parameter), if it's possible.

typescript/src/schema/proto.ts
```diff
@@ -288,6 +289,16 @@ function toLRInterface(type: string, inputType: string) {
   return inputType.replace(/\.([^.]+)$/, '.I' + type);
 }
 
+export function getHeaderParms(rule: plugin.google.api.IHttpRule): string[] {
+  const message =
+    rule.post || rule.delete || rule.get || rule.put || rule.patch;
+  if (message) {
+    const res = message.match(/{(.*?)(?==)/);
```
alexander-fenster: No need to have `(?=  )` around the final `=`. The meaning of `(?=  )` is "group together but don't extract to a match group". Since you only have one character `=` inside `(?=  )`, there is no need to put it into any kind of brackets if you don't need to extract them.

Just `/{(.*?)=/` should work.

typescript/src/schema/proto.ts
```diff
@@ -288,6 +289,16 @@ function toLRInterface(type: string, inputType: string) {
   return inputType.replace(/\.([^.]+)$/, '.I' + type);
 }
 
+export function getHeaderParms(rule: plugin.google.api.IHttpRule): string[] {
```
alexander-fenster: `Params`, not `Parms` :)

typescript/src/schema/proto.ts
```diff
@@ -382,6 +393,10 @@ function augmentMethod(
   if (method.methodConfig.timeout) {
     method.timeoutMillis = milliseconds(method.methodConfig.timeout);
   }
+  if (method.options && method.options['.google.api.http']) {
+    const httpRule = method.options['.google.api.http'];
+    method.headerRequestParams = getHeaderParms(httpRule);
```
alexander-fenster: it means that if a method does not have `.google.api.http` annotation the `.headerRequestParam` will be `undefined`, right? Let's explicitly assign it to `[]`.

typescript/src/util.ts
```diff
@@ -105,3 +105,11 @@ String.prototype.toSnakeCase = function(this: string): string {
   }
   return words.join('_');
 };
+
+Array.prototype.camelCaseBeforeDot = function(this: string[]): string {
+  if (this.length <= 1) {
+    return this.toString().toCamelCase();
+  }
+  const res = this.slice(0, -1).map(w => w.toCamelCase());
+  return res.join('!.') + '!.' + this[this.length - 1];
```
alexander-fenster: Wait. `!.` is a TypeScript specific, we must not have it in code :)  There should be a `for` loop in the templates doing that.

alexander-fenster: Alternatively, if you want to leave it here, pass a string to join as a parameter:
```ts
Array.prototype.camelCaseBeforeDot = function(this: string[], joiner: string): string { ... }
```

then you will call it from a template:
```ts
arr.camelCaseBeforeDot("!.")
```

and it will be fine since the function won't have any TypeScript specific code.

typescript/test/unit/proto.ts
```diff
@@ -15,8 +15,27 @@
 import * as assert from 'assert';
 
 import * as plugin from '../../../pbjs-genfiles/plugin';
-import { Proto } from '../../src/schema/proto';
+import { getHeaderParms } from '../../src/schema/proto';
 
 describe('schema/proto.ts', () => {
-  // TODO: test service augmentation
+  it('get header parameters from http rule', () => {
```
alexander-fenster: nit: it's a tradition to start test descrptions with "should":

```js
it('should get header parameters from http rule', () => { ... });
```

that way it reads "it should ..." like as if it is written in English.

alexander-fenster: ...or just `it('gets header parameters from http rule')`, `it('works with no parameters')`, etc. - just make it a real sentence.

templates/typescript_gapic/src/$version/$service_client.ts.njk
xiaozhenliu-gg5: sounds great! but do you mean unit test? we have` keymanager ` `translate`, `text-to-speech`API which respects this routing header, do we want to add one more baseline test, or just test this functionality in the `unit/`? Thanks!

typescript/src/schema/proto.ts
xiaozhenliu-gg5: Cool, just realize that for `/v1/{crypto_key_version.name=projects/*/locations/*/keyRings/*/cryptoKeys/*/cryptoKeyVersions/`

`/{(.*?)=/` will match `[{crypto_key_version.name=, crypto_key_version.name]`
and `/{(.*?)(?==)/` will match `[{crypto_key_version.name, crypto_key_version.name]` and I and taking the first element which will give the wrong answer of `crypto_key_version.name=`.

But with choosing the second element, it works perfect

typescript/src/schema/proto.ts
xiaozhenliu-gg5: Got it!

typescript/src/schema/proto.ts
xiaozhenliu-gg5: That makes sense to me, since it will check the length of `headerRequestParam` array

typescript/test/unit/proto.ts
xiaozhenliu-gg5: Sounds reasonable, made changes

templates/typescript_gapic/test/gapic-$service-$version.ts.njk
xiaozhenliu-gg5: sure! put `initRequestWithHeaderParam` method in macro of `_util.njk` and import it

templates/typescript_gapic/src/$version/$service_client.ts.njk
alexander-fenster: Maybe add `logging`? (in a separate PR)

templates/typescript_gapic/_util.njk
```diff
@@ -0,0 +1,11 @@
...
+            const request: protosTypes{{ method.inputInterface }} = {};
+{%- if method.headerRequestParams.length > 1 %}
+{%- set chain = "request" -%}
+{%- for field in method.headerRequestParams.slice(0, -1) %}
+            {{ chain }}.{{ field.toCamelCase() }} = {};
+{%- set chain = chain + "." + field.toCamelCase() -%}
+{%- endfor %}
+            {{ chain }}.{{ method.headerRequestParams.slice(-1)[0] }} = '';
+{%- endif %}
+{%- endmacro -%}
```
alexander-fenster: last nit!  please add a new line here :)

typescript/src/schema/proto.ts
```diff
@@ -2,6 +2,7 @@ import * as plugin from '../../../pbjs-genfiles/plugin';
 import { CommentsMap } from './comments';
 import * as objectHash from 'object-hash';
 import { milliseconds } from '../util';
+import { METHODS } from 'http';
```
alexander-fenster: unused import?

PR #94, opened 2019-11-01
feat: add jsdoc comment for parameters

### LINKED ISSUE #84: jsdoc
All those `@param`, `@returns` stuff for each of the generated methods.

xiaozhenliu-gg5:
fix #84 

we have a commentsMap in file level whose structure is ` <name: string, comment: [paramName: string, paramType: string, commentLines:string[]]>`
service: <serviceName, comment>
method: <serviceName : methodName, comment>
param: <messageName: fieldName, comment>
For `paramName` & `paramType` they are only set if it's for a parameter, and left "" for method & service.

1. Add util function for printing comments of method
2. Construct commentsMap in proto constructor and get comments for each level in augment function like `augmentService` `augmentMethod`
3. update baseline test

### DIFF
--- a/templates/typescript_gapic/_util.njk
+++ b/templates/typescript_gapic/_util.njk
@@ -1,0 +1,11 @@
+{%- macro initRequestWithHeaderParam(method) -%}
+            const request: protosTypes{{ method.inputInterface }} = {};
+{%- if method.headerRequestParams.length > 1 %}
+{%- set chain = "request" -%}
+{%- for field in method.headerRequestParams.slice(0, -1) %}
+            {{ chain }}.{{ field.toCamelCase() }} = {};
+{%- set chain = chain + "." + field.toCamelCase() -%}
+{%- endfor %}
+            {{ chain }}.{{ method.headerRequestParams.slice(-1)[0] }} = '';
+{%- endif %}
+{%- endmacro -%}
--- a/templates/typescript_gapic/_util.njk
+++ b/templates/typescript_gapic/_util.njk
@@ -1,3 +1,20 @@
+{#-
+
+Copyright 2019 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+-#}
 {%- macro initRequestWithHeaderParam(method) -%}
             const request: protosTypes{{ method.inputInterface }} = {};
 {%- if method.headerRequestParams.length > 1 %}
--- a/templates/typescript_gapic/_util.njk
+++ b/templates/typescript_gapic/_util.njk
@@ -13,7 +13,12 @@
 {%- macro printCommentsForParams(method) -%}
 {%- set commentsMap = method.paramComment -%}
 {%- for oneComment in commentsMap %}
- * @param { {{- oneComment.paramType.match('_([^_]+)$')[1].toLowerCase() -}} } request.{{ oneComment.paramName }}
+{%- set type = oneComment.paramType %}
+{%- if type.startsWith('.') %}
+ * @param { {{- type.substring(1) -}} } request.{{ oneComment.paramName }}
+{%- else %}
+ * @param { {{- convertParamType(oneComment.paramType) -}} } request.{{ oneComment.paramName }}
+{%- endif %}
 {%- set lines = oneComment.comments -%}
 {%- for line in lines %} 
  * {{ line.replaceAll('*/', '* /') }} 
@@ -26,4 +31,22 @@
 {%- for line in lines %}
  * {{ line }}
 {%- endfor %}
+{%- endmacro -%}
+
+{%- macro printRequest() %}
+ * @param {Object} request
+ *  The request object that will be sent.
+{%- endmacro -%}
+
+{%- macro convertParamType(paramType) -%}
+{%- if paramType.includes('TYPE_BYTES') %}
+{%- set type = paramType.replace('TYPE_BYTES', 'Buffer') %}
+{%- elif paramType.includes('TYPE_BOOL') %}
+{%- set type = paramType.replace('TYPE_BOOL', 'boolean') %}
+{%- elif paramType.includes('TYPE_STRING') %}
+{%- set type = paramType.replace('TYPE_STRING', 'string') %}
+{%- else %}
+{%- set type = 'number' %}
+{%- endif %}
+          {{- type -}}
 {%- endmacro -%}
--- a/templates/typescript_gapic/_util.njk
+++ b/templates/typescript_gapic/_util.njk
@@ -13,7 +13,7 @@
 {%- macro printCommentsForParams(method) -%}
 {%- set commentsMap = method.paramComment -%}
 {%- for oneComment in commentsMap %}
- * @param { {{ oneComment.paramType }} } request.{{ oneComment.paramName }}
+ * @param { {{- oneComment.paramType.match('_([^_]+)$')[1].toLowerCase() -}} } request.{{ oneComment.paramName }}
 {%- set lines = oneComment.comments -%}
 {%- for line in lines %} 
  * {{ line.replace('*/', '* /') }} 
--- a/templates/typescript_gapic/_util.njk
+++ b/templates/typescript_gapic/_util.njk
@@ -13,10 +13,10 @@
 {%- macro printCommentsForParams(method) -%}
 {%- set commentsMap = method.paramComment -%}
 {%- for oneComment in commentsMap %}
-  * @param { {{ oneComment.paramType }} } request.{{ oneComment.paramName }}
+ * @param { {{ oneComment.paramType }} } request.{{ oneComment.paramName }}
 {%- set lines = oneComment.comments -%}
 {%- for line in lines %} 
-  * {{ line }} 
+ * {{ line }} 
 {%- endfor %}
 {%- endfor %}
 {%- endmacro -%}
@@ -24,6 +24,6 @@
 {%- macro printCommentsForMethod(method) -%}
 {%- set lines = method.comments -%}
 {%- for line in lines %}
-  * {{ line }}
+ * {{ line }}
 {%- endfor %}
 {%- endmacro -%}
--- a/templates/typescript_gapic/_util.njk
+++ b/templates/typescript_gapic/_util.njk
@@ -63,9 +63,7 @@ limitations under the License.
 {%- endmacro -%}
 
 {%- macro printReturn(method, service) -%}
-{%- if method in service.simpleMethods %}
-{{- printReturnSimpleMethod(method) }}
-{%- elif method.longRunning %}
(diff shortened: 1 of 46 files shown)


xiaozhenliu-gg5:
Questions: 
1. what do we expect for the type of parameter? Now I'm printing the `field.type` while seems it's not quite same with the generated client lib.

2. There are cases that (like fail test ) ` /* ......... project*/ location /*....... */` in the comment which will be interpreted as the break of two comments, can you give some suggestion how do we handle that? Thanks!

typescript/test/testdata/texttospeech/src/v1/text_to_speech_client.ts.baseline
```diff
@@ -232,9 +232,20 @@ export class TextToSpeechClient {
   // -------------------
   // -- Service calls --
   // -------------------
-  /*
-  Returns a list of Voice supported for synthesis. 
-  */
+/**
+ *  Returns a list of Voice supported for synthesis.
+ * 
+ * @param { TYPE_STRING } request.language_code
```
alexander-fenster: Looks like we need to have a mapping from `TYPE_*` to actual JavaScript types.

alexander-fenster:
Thanks, that looks great! It's LGTM but we want to have actual types in `@param` messages. Let's try to figure out what are possible values for those types, and make a map or a macro (preferably in templates if possible, since it's obviously specific to TypeScript / JavaScript).

typescript/test/testdata/texttospeech/src/v1/text_to_speech_client.ts.baseline
xiaozhenliu-gg5: Got it!

xiaozhenliu-gg5:
I checked the kms client library and found that gapic-gen add one space between `* /` to escape the special case. I think it's a great way out for our micro-gen as well. :)

typescript/test/testdata/showcase/src/v1beta1/echo_client.ts.baseline
```diff
@@ -430,10 +467,20 @@ export class EchoClient {
...
- expanded words, this method returns a paged list of expanded words. 
-  */
+/**
+ *  This is similar to the Expand method but instead of returning a stream of
+ *  expanded words, this method returns a paged list of expanded words.
+ * 
+ * @param {string} request.content 
+ *  The string to expand. 
+ * 
+ * @param {int32} request.page_size
```
alexander-fenster: That won't work :(  There is no `int32` type in JavaScript. We need to have a real mapping.

The valid types are `object`, `number`, and `string`. If possible, it would be _great_ to use an actual message type instead of `object` (since we know it), but it can go into a separate PR if you prefer.

typescript/test/testdata/translate/src/v3beta1/translation_service_client.ts.baseline
```diff
@@ -304,9 +304,80 @@ export class TranslationServiceClient {
   // -------------------
   // -- Service calls --
   // -------------------
-  /*
-  Translates input text and returns translated text. 
-  */
+/**
+ *  Translates input text and returns translated text.
+ * 
+ * @param {string} request.contents
```
alexander-fenster: This should be `string[]` since it's an array.

alexander-fenster:
Example of the mapping I'm talking about, from gapic-generator: https://github.com/googleapis/gapic-generator/blob/4524f2d025adee7998bc6dd193f4edffaa969b1c/src/main/java/com/google/api/codegen/transformer/nodejs/NodeJSModelTypeNameConverter.java

typescript/test/testdata/showcase/src/v1beta1/echo_client.ts.baseline
alexander-fenster: upd. also `boolean` and `Buffer`

xiaozhenliu-gg5: Oh, I see, you were talking about the primitive type of js, working on this, thanks!

typescript/test/testdata/showcase/src/v1beta1/echo_client.ts.baseline
```diff
@@ -66,10 +66,7 @@ export interface PaginationResponse<
 
 export class EchoClient {
   /*
-  This service is used showcase the four main types of rpcs - unary, server
- side streaming, client side streaming, and bidirectional streaming. This
- service also exposes methods that explicitly implement server delay, and
- paginated calls.
+   This service is used showcase the four main types of rpcs - unary, server, side streaming, client side streaming, and bidirectional streaming. This, service also exposes methods that explicitly implement server delay, and, paginated calls.,
```
alexander-fenster: Something is wrong here, it was previously properly split into lines and now it's just one long line.

Can we keep splitting it, and also add `*` at the start of each line?

typescript/test/testdata/showcase/src/v1beta1/echo_client.ts.baseline
```diff
@@ -287,9 +284,19 @@ export class EchoClient {
...
   // -------------------
-  /*
-  This method simply echos the request. This method is showcases unary rpcs. 
-  */
+/**
+ * @param {Object} request
+ *  The request object that will be sent.
+ * @param {string} request.content 
+ *  The content to be echoed by the server. 
+ *
```
alexander-fenster: If possible, we should not have empty lines between `@param` lines.

typescript/test/testdata/texttospeech/src/v1/text_to_speech_client.ts.baseline
```diff
@@ -50,7 +50,7 @@ export interface Callback<
 
 export class TextToSpeechClient {
   /*
-  Service that implements Google Cloud Text-to-Speech API.
+   Service that implements Google Cloud Text-to-Speech API.,
```
alexander-fenster: This is weird, why did it add a comma?

alexander-fenster: I actually see this comma added in all baselines. Seems like something is wrong with a template :)

xiaozhenliu-gg5: yes, you are right, I directly print an array of comments which should be split and printed, thanks!

typescript/test/testdata/showcase/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: Sure, made changes. The data structure for service.comments changed, so it's acting wired now :)

typescript/test/testdata/showcase/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: Of course, just found that it's not from template, but from comments in proto files, they all have an empty line by the end of each comments of parameter.

templates/typescript_gapic/_util.njk
```diff
@@ -9,3 +9,57 @@
...
+{%- endmacro -%}
+
+{%- macro printRequest() %}
+ * @param {Object} request
+ *   The request object that will be sent.
+{%- endmacro -%}
+
+{%- macro printReturn() %}
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [LogMetric]{@link google.logging.v2.LogMetric}.
```
alexander-fenster: Oh no! There should not be any `LogMetric` here, it should take a parameter.

Also, please notice that different type of calls (unary, streaming, paging, long running) have different `@return` descriptions. It's OK to have separate macros for all of them, or do `if longrunning ... else if paging ...` etc.

alexander-fenster: I think this is the last comment here, everything else looks great!

xiaozhenliu-gg5: Thanks! Checked gapic-gen seems for @return part, there's also returnTypeName & return leading comments logic for printing, just like parameters. It seems not an easy fix, but let me explore more. :)

templates/typescript_gapic/_util.njk
```diff
@@ -9,3 +9,104 @@
...
+ *   The first element of the array is an object representing [{{- toMessageName(method.outputType) -}}]{@link {{ method.outputType.substring(1) }}}.
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+{%- endmacro -%}
+
+{%- macro printReturnPagingServerMethod(method) %}
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [{{- toMessageName(method.outputType) -}}]{@link {{ method.outputType.substring(1) }}}.
+ *   
+ *   When autoPaginate: false is specified through options, the array has three elements.
+ *   The first element is Array of [EchoResponse]{@link google.showcase.v1beta1.EchoResponse} in a single response.
```
alexander-fenster: `EchoResponse` should not be here :)

templates/typescript_gapic/_util.njk
```diff
@@ -9,3 +9,104 @@
...
+{%- endfor %}
+{%- endmacro -%}
+
+{%- macro printRequest() %}
+ * @param {Object} request
+ *   The request object that will be sent.
+{%- endmacro -%}
+
+{%- macro printReturn(method, service) -%}
+{%- if method in service.simpleMethods %}
```
alexander-fenster: Let's move it down. If `method` is not long running, paging, etc. then just `else` it's simple.

templates/typescript_gapic/_util.njk
```diff
@@ -9,3 +9,104 @@
...
+
+{%- macro printReturnPagingServerMethod(method) %}
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [{{- toMessageName(method.outputType) -}}]{@link {{ method.outputType.substring(1) }}}.
+ *   
+ *   When autoPaginate: false is specified through options, the array has three elements.
+ *   The first element is Array of [EchoResponse]{@link google.showcase.v1beta1.EchoResponse} in a single response.
+ *   The second element is the next request object if the response
+ *   indicates the next page exists, or null. The third element is
+ *   an object representing [PagedExpandResponse]{@link google.showcase.v1beta1.PagedExpandResponse}.
```
alexander-fenster: same: `PagedExpandResponse` should not be here

templates/typescript_gapic/_util.njk
```diff
@@ -9,3 +9,104 @@
...
+{%- macro printReturnClientStreamingMethod(method) %}
+ * @returns {Stream} - A writable stream which accepts objects representing 
+ * [{{- toMessageName(method.outputType) -}}]{@link {{ method.outputType.substring(1) }}}.
+{%- endmacro -%}
+
+{%- macro printReturnBidiStreamingMethod(method) %}
+ * @returns {Stream} 
+ *   An object stream which is both readable and writable. It accepts objects
+ *   representing [[{{- toMessageName(method.outputType) -}}]{@link {{ method.outputType.substring(1) }}} for write() method, and
+ *   will emit objects representing [EchoResponse]{@link google.showcase.v1beta1.EchoResponse} on 'data' event asynchronously.
```
alexander-fenster: `EchoResponse`

templates/typescript_gapic/_util.njk
xiaozhenliu-gg5: Oh! nice catch!

templates/typescript_gapic/_util.njk
xiaozhenliu-gg5: Sure!

PR #96, opened 2019-11-01
docs: a little bit of README

alexander-fenster:
Adding some very basic `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,88 @@
 # API Client Generator for TypeScript
 
-**This is not an official Google product.**
+## Background
 
-This repository will contain the source code of a client library generator
-for APIs specified by [protocol
-buffers](https://developers.google.com/protocol-buffers/),
-such as those inside Google. It takes a protocol buffer (with particular
-annotations) and uses it to generate a client library in TypeScript.
+This tool is a TypeScript client library generator for network APIs specified
+by [protocol
+buffers](https://developers.google.com/protocol-buffers/) (including, but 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,88 @@
...
+
+### Run the generator to generate TypeScript libraries
+
+The easiest way to get started is to use our Docker image:
+
+```sh
+$ mkdir showcase-typescript
+$ docker run --rm \
+  --mount type=bind,source=`pwd`/google/showcase/v1beta1,destination=/in/google/showcase/v1beta1,readonly \
+  --mount type=bind,source=`pwd`/showcase-typescript,destination=/out \
```
xiaozhenliu-gg5: should we give users the place to download the docker image? Or we will add it later, thanks!

README.md
```diff
@@ -1,10 +1,88 @@
...
+### Write your JavaScript or TypeScript code that uses the server!
+
+Assuming you have Node.js installed and `node` binary in your path:
+
+```sh
+$ node
+Welcome to Node.js v12.1.0.
+Type ".help" for more information.
+> const showcase = require('.')  // assuming you're in showcase-typescript
+undefined
```
xiaozhenliu-gg5: is this `undefined` intentional?

alexander-fenster: Yes, I'm just showing verbatim Node.js output.

README.md
alexander-fenster: That's exactly the command that should be used, according to [AIP 4290](https://aip.dev/4290). There is no need to download Docker image, `docker run` will pull the image from gcr.io.

PR #97, opened 2019-11-01
fix: add proper warning message

xiaozhenliu-gg5:


alexander-fenster:
Since it's a code change, it's a `fix:` :)  Chore is something like fixing CircleCI config.

PR #99, opened 2019-11-04
fix: no ts-ignore in generated code

alexander-fenster:
1. `@ts-ignore` is bad  (`any` is also bad but better). Sooner or later we'll figure out how to get rid of that `any` as well.

2. To catch the problem that happened with `cloudbuild` (stopped compiling after `gts fix`), let's try compiling the generating code once more after fixing its formatting in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,7 @@ jobs:
             npm install
             npm test
             npm run fix
+            npm run compile
       - run:
           name: Run unit tests and gts fix of the generated KMS library
           command: |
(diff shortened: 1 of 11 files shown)


alexander-fenster:
Yes, because it's exactly what happened in https://github.com/googleapis/nodejs-cloudbuild/pull/15. `npx gts fix` moved stuff around so the `// @ts-ignore` line stopped being attached to the string it was originally followed by. That's of course a `gts` bug but it still breaks our code.

PR #101, opened 2019-11-04
chore: add licenses to all files

alexander-fenster:
Adding Apache 2.0 license to all our source files.

@bcoe Can we please have a license checker bot enabled on this repo? Thank you!


alexander-fenster:
@xiaozhenliu-gg5 The rule is to put a license into all source files. It's arguable if `.njk` files are source files or not, but since it's easy to add the license as a comment, let's just do it.

PR #102, opened 2019-11-04
chore: move string manipulation methods to template

### LINKED ISSUE #95: toLRInterface and toInterface should not exist in the code
I just realized these two functions in `proto.ts` actually depend on JavaScript `protobufjs` feature (interfaces). In the perfect world, they must not exist in the code, but should be in templates if possible. We are doing great job keeping the generator as general purpose as possible so it would be great to refactor this.

Rationale: if we ever decide to generate another language by substituting another set of templates, these `toLRInterface` and `toInterface` will be of no use, and the contents of the augmented method object won't fit the target language.

xiaozhenliu-gg5:
fix #95

### DIFF
--- a/templates/typescript_gapic/_util.njk
+++ b/templates/typescript_gapic/_util.njk
@@ -9,3 +9,8 @@
             {{ chain }}.{{ method.headerRequestParams.slice(-1)[0] }} = '';
 {%- endif %}
 {%- endmacro -%}
+
+{%- macro toInterface(inputType) -%}
+{%- set inputType = inputType | replace('\.([^.]+)$', '.I$1') -%}
+    {{ inputType }}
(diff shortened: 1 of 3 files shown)


xiaozhenliu-gg5:
The `string.replace('regex', '.I$1')` does not work in the template, not sure why, but I changed to divide and combine the string instead. Let me know if you have better way, thanks!

xiaozhenliu-gg5:
By the way, the update baseline script saves me a lot of time, thank you for making this simpler!

alexander-fenster:
@xiaozhenliu-gg5 I made a manual merge but seems like the comments code should be modified now :)  Can you please take a look? Just `git pull` from this branch and `npm run compile` to see the problem.

PR #103, opened 2019-11-05
feat: add docker test script

xiaozhenliu-gg5:


docker/test.sh
```diff
@@ -0,0 +1,22 @@
...
+# Create new directory showcase-typescript. 
+mkdir showcase-typescript
+# Pull the docker image
+docker pull gcr.io/gapic-images/gapic-generator-typescript:latest
+# Use Docker Image for generating showcase client library
+docker run --rm \
+  --mount type=bind,source=`pwd`/typescript/test/protos/google/showcase/v1beta1,destination=/in/typescript/test/protos/google/showcase/v1beta1,readonly \
+  --mount type=bind,source=`pwd`/showcase-typescript,destination=/out \
+  gcr.io/gapic-images/gapic-generator-typescript:latest
+# Test geenrated client library
```
alexander-fenster: `generated`

xiaozhenliu-gg5: Thanks!

docker/test.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+### Test script pulling the docker image and use it against showcase proto.
+
+# Docker image tag: gapic-generator-typescript:latest.
+# Create new directory showcase-typescript. 
+mkdir showcase-typescript
+# Pull the docker image
+docker pull gcr.io/gapic-images/gapic-generator-typescript:latest
```
alexander-fenster: Here it makes sense to test local `gapic-generator-typescript` image before pushing to `gcr.io`, and not after. So don't pull here, just do `docker run ........ gapic-generator-typescript`. If it works, we do `docker push`.

docker/test.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+### Test script pulling the docker image and use it against showcase proto.
+
+# Docker image tag: gapic-generator-typescript:latest.
+# Create new directory showcase-typescript. 
+mkdir showcase-typescript
```
alexander-fenster: It will fail if it already exists. Let's do `rm -rf .showcase-typescript` first, then `mkdir`. Also, it makes sense to put its name into a variable since it's used more than once.

alexander-fenster: Also, I think the dot is missing (should be `.showcase-typescript`, as it's in `.gitignore`).

docker/test.sh
```diff
@@ -0,0 +1,22 @@
...
+  --mount type=bind,source=`pwd`/showcase-typescript,destination=/out \
+  gcr.io/gapic-images/gapic-generator-typescript:latest
+# Test geenrated client library
+cd showcase-typescript
+npm install  # install dependencies
+npm run fix  # format the code
+npm test     # run unit tests
+
+# Test succeed
+echo 'docker test succedd!'
```
alexander-fenster: `succeeded`

docker/test.sh
xiaozhenliu-gg5: Sure! makes sense to me!

docker/test.sh
xiaozhenliu-gg5: Thanks!

docker/test.sh
xiaozhenliu-gg5: :( sorry about the typo

docker/test.sh
alexander-fenster: `showcase-typescript` occurs three times in this script, let's have a variable `OUTPUT_DIR=.showcase_typescript`. That way it would be much easier to change the location if needed.  Also, please note that the `.gitignore` says `.showcase-typescript`  (with a dot).

alexander-fenster: Actually, it just looks you didn't push your latest changes :)

xiaozhenliu-gg5: sorry!

docker/test.sh
```diff
@@ -0,0 +1,23 @@
...
+# Docker image tag: gapic-generator-typescript:latest.
+DIR_NAME=.showcase-typescript
+# Remove test directory if it already exists
+rm -rf $DIR_NAME
+# Create new directory showcase-typescript. 
+mkdir $DIR_NAME
+# Use Docker Image for generating showcase client library
+docker run --rm \
+  --mount type=bind,source=`pwd`/typescript/test/protos/google/showcase/v1beta1,destination=/in/typescript/test/protos/google/showcase/v1beta1,readonly \
+  --mount type=bind,source=`pwd`/.showcase-typescript,destination=/out \
```
alexander-fenster: `$DIRNAME` here :)

docker/test.sh
```diff
@@ -0,0 +1,23 @@
...
+  gapic-generator-typescript:latest
+# Test generated client library
+cd .showcase-typescript
```
alexander-fenster: and here

PR #108, opened 2019-11-06
chore: add options to comments for all methods

xiaozhenliu-gg5:
fix #108

### DIFF
--- a/typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
+++ b/typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
@@ -281,6 +281,8 @@ export class KeyManagementServiceClient {
  *
  * @param {Object} request
  *   The request object that will be sent.
+ * @param {object} [options] - The configuration object. See the subsequent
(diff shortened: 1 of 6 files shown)


typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
```diff
@@ -281,6 +281,8 @@ export class KeyManagementServiceClient {
  *
  * @param {Object} request
  *   The request object that will be sent.
+ * @param {object} [options] - The configuration object. See the subsequent
```
alexander-fenster: Let's remove `the subsequent parameters` thing since there are no "subsequent parameter". I would better replace the whole text with this:

```
 * @param {Object} [options]
     Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
```

xiaozhenliu-gg5: Cool, updating the text, thanks!

PR #110, opened 2019-11-06
chore: move js doc around

xiaozhenliu-gg5:
Checked the compiled js client file does have all proper comments now.


templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -371,6 +364,12 @@ export class {{ service.name }}Client {
           protosTypes{{ util.toInterface(method.outputInterface) }},
           protosTypes{{ util.toInterface(method.inputInterface) }}|undefined,
           {}|undefined>): void;
+/**
```
alexander-fenster: I'm starting to think it makes perfect sense to have one `util` function combining these four lines :)

xiaozhenliu-gg5: Sure, that's a great idea!

PR #116, opened 2019-11-08
feat: add comments for page stream method

### LINKED ISSUE #112: streaming interface for paging calls
Who knew we had this?

https://github.com/googleapis/gapic-generator/blob/e3501faea84f61be2f59bd49a7740a486a02fa6b/src/main/resources/com/google/api/codegen/nodejs/main.snip#L477-L508

Since the generated code does not have unit tests for this, we can also skip testing it, but it makes sense to try adding it to the test showcase applications (both JS and TS).

xiaozhenliu-gg5:
Add function for page streaming comments and restructure util file

fix #112

### DIFF
--- a/typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
+++ b/typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
@@ -1633,6 +1633,7 @@ export class KeyManagementServiceClient {
     });
     return this._innerApiCalls.listKeyRings(request, options, callback);
   } 
+
 /**
  * Equivalent to {@link listKeyRings}, but returns a NodeJS Stream object.
  *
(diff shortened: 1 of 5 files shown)


typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
```diff
@@ -1731,7 +1766,44 @@ export class KeyManagementServiceClient {
     });
     return this._innerApiCalls.listCryptoKeys(request, options, callback);
   } 
-
+/**
```
alexander-fenster: nit: can you add an empty line before this comment block?

xiaozhenliu-gg5: sure!

PR #117, opened 2019-11-08
chore: optional parameter should be in square brackets and camel case

xiaozhenliu-gg5:
1. optional parameter should be in square brackets
2. all parameters are in camel case.

### DIFF
--- a/typescript/src/schema/comments.ts
+++ b/typescript/src/schema/comments.ts
@@ -20,8 +20,8 @@ export interface Comment {
   paramName: string;
   paramType: string;
   comments: string[];
-  // LABEL_OPTIONAL = 1, LABEL_REQUIRED = 2, LABEL_REPEATED = 3
-  label?: plugin.google.protobuf.FieldDescriptorProto.Label;
+  // LABEL_OPTIONAL = 1, LABEL_REQUIRED = 2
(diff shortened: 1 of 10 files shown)


typescript/src/schema/comments.ts
```diff
@@ -121,13 +123,15 @@ export class CommentsMap {
                 ) {
                   paramType = field.typeName!;
                 }
+                const label = field.label ? field.label : undefined;
```
alexander-fenster: Why not just `const label = field.label`? If it's undefined, it will be undefined.

typescript/test/testdata/texttospeech/src/v1/text_to_speech_client.ts.baseline
```diff
@@ -318,11 +318,11 @@ export class TextToSpeechClient {
  *
  * @param {Object} request
  *   The request object that will be sent.
- * @param {google.cloud.texttospeech.v1.SynthesisInput} request.input 
+ * @param {google.cloud.texttospeech.v1.SynthesisInput} [request.input]
```
alexander-fenster: Something is going wrong. This field must be marked as required in proto, why is it considered optional here?

PR #118, opened 2019-11-08
fix: require @types/long

alexander-fenster:
This does not look like a 100% clean solution but I checked that pub/sub does exactly this https://github.com/googleapis/nodejs-pubsub/blob/master/package.json#L60.

The generated `protos.d.ts` uses `Long` type and we must do so as well to make sure all types are imported properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "gapic-generator-typescript": "build/src/start_script.js"
   },
   "scripts": {
+    "baseline": "node build/tools/update-baselines.js",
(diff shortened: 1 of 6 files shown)


JustinBeckwith:
You're going to hate this, but can we add it as a runtime dependency of gax?  That way we don't need to go add this dependency and keep it up to date in ~80 repos.

alexander-fenster:
That's actually a good idea - will do!

PR #127, opened 2019-11-11
fix: separate jsdoc with api version

xiaozhenliu-gg5:
run `npm run docs` in generated client library, all jsdocs in two apis are combined into one place
```
TextToSpeechClient
- method1 (from v1)
- method1 (from v1beta1)
- method2 
....
````

After fix, the layout will be
```
v1.TextToSpeechClient
- method1 
- method2
....

v1beta1.TextToSpeechClient
- method1 
- method2 
....
````

### CODE REVIEW

**Comment by xiaozhenliu-gg5** [comment]
Questions: 
Inconsistency between googlapis.dev/nodejs/text-to-speech and our jsdoc
```
synthesizeSpeech(request, optionsopt, callbackopt) → {Promise}
```
and now we have 
```
synthesizeSpeech(request, optionsopt) → {Promise}
```

it's because of we don't have `@param callback` in comments, so we want to add it?

alexander-fenster:
No, we don't want to advertise the callback interface. This change is OK. We want people to use promises and `async`/`await`, we don't want to show callbacks in our documentation at all.

xiaozhenliu-gg5:
Thanks Alex, merge this PR

PR #128, opened 2019-11-12
feat: generate .jsdoc.js for client library

### LINKED ISSUE #92: generate .jsdoc.js
Should be very straightforward, I think it's just constant and we can copy it from any library, maybe with slight changes.

xiaozhenliu-gg5:
fix #92 
Tested showcase locally, which is working correctly.


templates/typescript_gapic/.jsdoc.js.njk
```diff
@@ -0,0 +1,46 @@
+/*!
+ * Copyright 2018 Google LLC. All Rights Reserved.
```
alexander-fenster: 2019

templates/typescript_gapic/package.json.njk
```diff
@@ -41,14 +41,17 @@ limitations under the License.
     "mv": "^2.1.1",
     "ncp": "^2.0.0",
     "tmp": "^0.1.0",
-    "typescript": "^3.7.0"
+    "typescript": "^3.7.0",
+    "jsdoc": "^3.5.5",
```
alexander-fenster: nit: can we try to keep the alphabetical order of the dependencies? `npm` does that.

templates/typescript_gapic/.jsdoc.js.njk
```diff
@@ -0,0 +1,46 @@
...
+    'plugins/markdown'
+  ],
+  source: {
+    include: [
+      'build/src'
+    ],
+    includePattern: '\\.js$'
+  },
+  templates: {
+    copyright: 'Copyright 2018 Google, LLC.',
```
alexander-fenster: 2019

templates/typescript_gapic/.jsdoc.js.njk
```diff
@@ -0,0 +1,46 @@
...
+    recurse: true,
+    verbose: true,
+    destination: './docs/'
+  },
+  plugins: [
+    'plugins/markdown'
+  ],
+  source: {
+    include: [
+      'build/src'
```
alexander-fenster: Can we look at `src` instead and include all `.ts`?  (I'm not sure it can work with TypeScript source, but can we try?)

JustinBeckwith: I am going to guess the parser will choke on TypeScript syntax, but it's worth trying.

typescript/test/testdata/translate/package.json.baseline
```diff
@@ -24,14 +24,17 @@
...
+    "jsdoc": "^3.5.5",
+    "jsdoc-fresh": "^1.0.1"
   },
   "scripts": {
     "lint": "gts check",
     "clean": "gts clean",
     "compile": "tsc -p . && cp -r protos build/",
     "compile-protos": "compileProtos src",
     "fix": "gts fix",
+    "docs": "jsdoc -c .jsdoc.js",
```
JustinBeckwith: Getting closer 😆 @alexander-fenster do we want to include linkinator and it's config, since we're using those everywhere as well?

alexander-fenster: Yes, makes sense. @xiaozhenliu-gg5 can you add `linkinator` to dependencies and scripts (like [here](https://github.com/googleapis/nodejs-pubsub/blob/master/package.json)), and its config like [this one](https://github.com/googleapis/nodejs-pubsub/blob/master/linkinator.config.json)?

alexander-fenster:
One more thing is that the old gapic-generator had the whole extra doc folder generated just for the types from protos. We are lucky to have `protos/protos.js` that has all the same information, with valid `jsdoc` comments. I think we can just include `protos/protos.js` into the jsdoc command to have all the proto types documented. @xiaozhenliu-gg5 can you try that?

typescript/test/testdata/translate/package.json.baseline
xiaozhenliu-gg5: Sure, added the dependency and tested it work

templates/typescript_gapic/.jsdoc.js.njk
xiaozhenliu-gg5: yes, it's unable to parse the ts file and throws error, not sure if there's other way to include `.ts` file.

templates/typescript_gapic/.jsdoc.js.njk
xiaozhenliu-gg5: thanks!

templates/typescript_gapic/package.json.njk
xiaozhenliu-gg5: sure, run `npx sort-package-json` to sort the dependencies, did not know there's a way to sort without extra package. Thanks!

templates/typescript_gapic/.jsdoc.js.njk
xiaozhenliu-gg5: thank you, modified it.

PR #136, opened 2019-11-14
fix: use pagination types from gax

alexander-fenster:
Using pagination types exported from `google-gax`.

CI won't work until gax v1.11.0 is released.


xiaozhenliu-gg5:
sorry I overlooked this interface when I moved common interfaces to gax, thanks for doing this!

PR #148, opened 2019-11-19
feat: let user pass package name to the generator

### LINKED ISSUE #134: accept an option for package name
We've got a weird problem when we started using `pack-n-play`. The generated system tests need to know the package name to `require` it properly, but the generator has no idea how the package is called.

E.g. for Text-to-speech, the actual name is `@google-cloud/text-to-speech` but we generate `require('texttospeech')` in system tests, and it fails.

For now, I solved this with a replace command in `synth.py`, but that's a hack. There are two possible options:

1. Make a generator flag to set the package name: `gapic-generator-typescript ..... --package-name @google-cloud/text-to-speech`

2. In `pack-n-play`, use some predefined name (e.g. `library`) for the library being tested, and write it to `package.json` manually, e.g. `"library": "file.tgz"` - that's what I did in my first system test (the one that is now removed).

I'm OK with both solutions (or any other possible solution you might have - let me know!), but I'd like to remove the `synth.py` hack as soon as possible :)

xiaozhenliu-gg5:
fix #134 

User can pass package name using optional `--package-name` argument to generator.

### DIFF
--- a/templates/typescript_gapic/package.json.njk
+++ b/templates/typescript_gapic/package.json.njk
@@ -16,7 +16,7 @@ limitations under the License.
 
 -#}
 {
-  "name": "{{ api.naming.productName.toKebabCase() }}",
+  "name": "{{ api.publishName}}",
   "version": "0.1.0",
   "description": "{{ api.naming.productName }} client for Node.js",
   "repository": "googleapis/nodejs-{{ api.naming.productName.toKebabCase() }}",
--- a/templates/typescript_gapic/package.json.njk
+++ b/templates/typescript_gapic/package.json.njk
@@ -16,7 +16,7 @@ limitations under the License.
 
 -#}
 {
-  "name": "{{ api.publishName}}",
+  "name": "{{ api.publishName }}",
   "version": "0.1.0",
   "description": "{{ api.naming.productName }} client for Node.js",
   "repository": "googleapis/nodejs-{{ api.naming.productName.toKebabCase() }}",
--- a/typescript/src/generator.ts
+++ b/typescript/src/generator.ts
@@ -23,7 +23,7 @@ import { API } from './schema/api';
 import { processTemplates } from './templater';
 import { commonPrefix, duration } from './util';
 
-export interface Map {
+export interface OptionsMap {
   [name: string]: string;
 }
 const readFile = util.promisify(fs.readFile);
@@ -43,6 +43,7 @@ export class Generator {
   request: plugin.google.protobuf.compiler.CodeGeneratorRequest;
   response: plugin.google.protobuf.compiler.CodeGeneratorResponse;
   grpcServiceConfig: plugin.grpc.service_config.ServiceConfig;
+  paramMap: OptionsMap;
   // This field is for users passing proper publish package name like @google-cloud/text-to-speech.
   publishName?: string;
 
@@ -50,6 +51,7 @@ export class Generator {
     this.request = plugin.google.protobuf.compiler.CodeGeneratorRequest.create();
     this.response = plugin.google.protobuf.compiler.CodeGeneratorResponse.create();
     this.grpcServiceConfig = plugin.grpc.service_config.ServiceConfig.create();
+    this.paramMap = {};
   }
 
   // Fixes gRPC service config to replace string google.protobuf.Duration
@@ -73,18 +75,16 @@ export class Generator {
 
   private getParamMap(parameter: string) {
     // Example: "grpc-service-config=texamplejson","package-name=packageName"
-    const paramMap: Map = {};
     const parameters = parameter.split(',');
     for (let param of parameters) {
       // remove double quote
       param = param.substring(1, param.length - 1);
       const arr = param.split('=');
-      paramMap[arr[0].toKebabCase()] = arr[1];
+      this.paramMap[arr[0].toKebabCase()] = arr[1];
     }
-    return paramMap;
   }
 
-  private async readGrpcServiceConfig(map: Map) {
+  private async readGrpcServiceConfig(map: OptionsMap) {
     if (map && map['grpc-service-config']) {
       const filename = map['grpc-service-config'];
       if (!fs.existsSync(filename)) {
@@ -99,7 +99,7 @@ export class Generator {
     }
   }
 
-  private async readPublishPackageName(map: Map) {
+  private async readPublishPackageName(map: OptionsMap) {
     if (map && map['package-name']) {
       this.publishName = map['package-name'];
     }
@@ -111,9 +111,9 @@ export class Generator {
       inputBuffer
     );
     if (this.request.parameter) {
-      const map = this.getParamMap(this.request.parameter);
-      await this.readGrpcServiceConfig(map);
-      await this.readPublishPackageName(map);
+      this.getParamMap(this.request.parameter);
+      await this.readGrpcServiceConfig(this.paramMap);
+      await this.readPublishPackageName(this.paramMap);
     }
   }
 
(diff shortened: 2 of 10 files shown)


templates/typescript_gapic/package.json.njk
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 -#}
 {
-  "name": "{{ api.naming.productName.toKebabCase() }}",
+  "name": "{{ api.publishName}}",
```
alexander-fenster: please add a space before `}}`

typescript/src/generator.ts
```diff
@@ -39,6 +39,8 @@ export class Generator {
   request: plugin.google.protobuf.compiler.CodeGeneratorRequest;
   response: plugin.google.protobuf.compiler.CodeGeneratorResponse;
   grpcServiceConfig: plugin.grpc.service_config.ServiceConfig;
+  //This field is for users passing proper module name for system test.
```
alexander-fenster: Please add a space after `//`

alexander-fenster: also, it's not just for system test.

typescript/src/generator.ts
```diff
@@ -66,7 +68,7 @@ export class Generator {
   }
 
   private async readGrpcServiceConfig(parameter: string) {
-    const match = parameter.match(/^["']?grpc-service-config=([^"]+)["']?$/);
+    const match = parameter.match(/^["']?grpc-service-config=([^"]+)["']?/);
```
alexander-fenster: if both `grpc-service-config` and `package-name` are passed, this won't work (the regex will take the whole string up to and including the `package-name` parameter).

I think at this point, since we have two parameters (and may have more), we need to parse it in some more accurate way. Let's make an assumption that parameter values cannot have spaces, then we can have a function that parses this string and returns a map:

```ts
function parseParameterString(parameters: string) {
  const params = parameters.split(' ');
  ...
  // return an object of key-value pairs
}
```

and then just parse it once and take `params["grpc-service-config"]`, `params["package-name"]`, etc.  What do you think?

typescript/src/schema/api.ts
```diff
@@ -32,19 +32,28 @@ export class API {
   hostName?: string;
   port?: string;
   mainServiceName?: string;
+  //This field is for users passing roper module name for system test.
```
alexander-fenster: `proper` (also, please add a space after `//`)

typescript/src/schema/api.ts
```diff
@@ -32,19 +32,28 @@ export class API {
...
+    publishName?: string
   ) {
     this.naming = new Naming(
       fileDescriptors.filter(
         fd => fd.package && fd.package.startsWith(packageName)
       )
     );
+    // users specify the actual package name, if not, set it to product name.
+    if (publishName) {
+      this.publishName = publishName;
```
alexander-fenster:
```ts
this.publishName = publishName ||  this.naming.productName.toKebabCase();
```

typescript/src/start_script.ts
```diff
@@ -34,6 +34,8 @@ const argv = yargs
   .describe('output_dir', 'Path to a directory for the generated code')
   .alias('grpc-service-config', 'grpc_service_config')
   .describe('grpc-service-config', 'Path to gRPC service config JSON')
+  .alias('package-name', 'package-name')
```
alexander-fenster: `.alias('package-name', 'package_name')` - note an underscore. The idea of those aliases is to allow both `--package-name` and `--package_name` because people tend to use both.

typescript/src/start_script.ts
```diff
@@ -71,6 +74,9 @@ if (grpcServiceConfig) {
     `--typescript_gapic_opt="grpc-service-config=${grpcServiceConfig}"`
   );
 }
+if (packageName) {
+  protocCommand.push(`--typescript_gapic_opt="package-name=${packageName}"`);
```
alexander-fenster: Hmmm. Will it actually concatenate options into one string if two instances of `--typescript-gapic-opt` are passed?  I was under impression we need to join them manually here.

typescript/src/generator.ts
xiaozhenliu-gg5: yes, I noticed the same thing that >(the **former** regex will take the whole string up to and including the package-name parameter). 
so I modified a little bit the regex, now match[1] only takes the correct argument `example.json` and `@google-cloud/text-to-speech`.  Test case `extra-option.ts` takes both `grpc-service-config` and `package-name` which works fine. 

But I think making key-value pair makes perfect sense for adding more options to generator.

typescript/src/start_script.ts
xiaozhenliu-gg5: I see, thanks for clarification!

typescript/src/start_script.ts
xiaozhenliu-gg5: just printed out the `this.request.parameter` for `CodeGeneratorRequest` and it gives:
```
"grpc-service-config=/Users/xiaozhenliu/gapic-generator-typescript/build/test/protos/google/cloud/texttospeech/v1/texttospeech_grpc_service_config.json","package-name=@google-cloud/text-to-speech"
```

Seems it has some methods to combine all `--typescript_gapic_opt` arguments together.

typescript/src/schema/api.ts
xiaozhenliu-gg5: Thanks, this is cleaner and more elegant!

typescript/src/schema/api.ts
xiaozhenliu-gg5: thank you!

typescript/src/start_script.ts
xiaozhenliu-gg5: when `CodeGenerateRequest.decode(inputBuffer)`, apart from fields `fileToGenerate`, `protoFile`, `compilerVersion` I think any other arguments get combined and go to `parameter`, which is one big string.

typescript/src/generator.ts
alexander-fenster: `^` at the beginning forces the grpc service config parameter to go first. maybe remove it?

typescript/src/schema/api.ts
alexander-fenster: still unresolved :)

typescript/src/start_script.ts
alexander-fenster: Then I really think it's better to split this string by a comma and have key value pairs.

typescript/src/generator.ts
xiaozhenliu-gg5: I will make a map for the parameters which makes easier to add more arguments, and more readable. thanks for the suggestion!

typescript/src/generator.ts
```diff
@@ -23,6 +23,9 @@ import { API } from './schema/api';
 import { processTemplates } from './templater';
 import { commonPrefix, duration } from './util';
 
+export interface Map {
```
alexander-fenster: JavaScript already has `Map`, let's call it `OptionsMap` or something like that.

typescript/src/generator.ts
```diff
@@ -65,22 +71,38 @@ export class Generator {
...
+      const arr = param.split('=');
+      paramMap[arr[0].toKebabCase()] = arr[1];
     }
-    const filename = match[1];
-    if (!fs.existsSync(filename)) {
-      throw new Error(`File ${filename} cannot be opened.`);
+    return paramMap;
+  }
+
+  private async readGrpcServiceConfig(map: Map) {
```
alexander-fenster: It makes sense to make this map a class member, and just access `this.map`.

typescript/src/schema/api.ts
alexander-fenster: still unresolved :) Still no space, and `roper`

typescript/src/generator.ts
xiaozhenliu-gg5: Sounds great

typescript/src/generator.ts
xiaozhenliu-gg5: sure!

typescript/src/schema/api.ts
xiaozhenliu-gg5: sorry, thought it was another file!

xiaozhenliu-gg5:
Thank you!

PR #150, opened 2019-11-20
fix: change style of comment in protos for protobufjs

alexander-fenster:
So this one is weird.

---

**TL;DR:** in this PR we convert all `//` comments to `/** ... */` in all the proto files that are copied to the generated client library.

---

In the monolith generator, we created the whole `src/v*/doc` folder (e.g. [here](https://github.com/googleapis/nodejs-datacatalog/tree/master/src/v1beta1/doc)) that was pretty much the copy of protos with comments, converted to JavaScript for `jsdoc`.

In the micro-generator, we don't do that. Instead, we run `pbjs` to generate `protos/protos.js` and then let `jsdoc` read that file, which just works.

But, I just realized that `jsdoc` does not consider `//` style comments from proto files and does not copy them to the resulting `protos/protos.js`.  Its documentation says [here](https://github.com/protobufjs/protobuf.js#pbjs-for-javascript):

> **ProTip!** Documenting your .proto files with `/** ... */`-blocks or (trailing) `/// ...` lines translates to generated static code.

So... we must convert the proto comments from `//` to `/** ... */` to make them go into the jsdoc. Luckily, we have a great place to do that - right in the code where we copy proto files into the resulting directory. This PR does exactly that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
     "test": "c8 --reporter=lcov mocha build/test/unit"
   },
   "dependencies": {
-    "file-system": "^2.2.2",
     "fs-extra": "^8.1.0",
     "get-stdin": "^7.0.0",
     "google-gax": "^1.9.0",
--- a/typescript/src/start_script.ts
+++ b/typescript/src/start_script.ts
@@ -18,7 +18,7 @@ import { execFileSync } from 'child_process';
 import * as path from 'path';
 import * as yargs from 'yargs';
 import * as fs from 'fs-extra';
-import {updateProtoComments} from './util';
+import { updateProtoComments } from './util';
 
 const googleGaxPath = path.dirname(require.resolve('google-gax')); // ...../google-gax/build/src
 const GOOGLE_GAX_PROTOS_DIR = path.join(googleGaxPath, '..', '..', 'protos');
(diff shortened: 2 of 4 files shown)


JustinBeckwith:
Yeah, this is weird 😆 Should we be proposing a change in `jsdoc` or `protobufjs` instead?  I'm a little worried about trying to accurately parse JavaScript in a way that's complete/effective here.

alexander-fenster:
I'm only parsing protos, not JavaScript, so it's _probably_ easier and more predictable.

`jsdoc` is out of question, it just reads what is in `protos/protos.js`.  But `protobufjs` did not put the comments there because they ignore `//`-style comments.

Fixing this in `pbjs` may be an option, but a harder one (let me say this - I spent like half an hour staring at that code and did not immediately figure out how to make it work there). So I still opt for local proto rewrite.

Note that we don't really use these protos for anything in the runtime (they are now published just for reference, we don't load them), and if any parsing problem happens, we'll catch it when `synthtool` fails (when we run `npx compileProtos src`).

alexander-fenster:
I need to think more about the proper way of doing this, given that we now can make protobuf.js releases.

PR #156, opened 2019-11-21
fix: correctly determine auto-paginated field

alexander-fenster:
When trying to convert redis in https://github.com/googleapis/nodejs-redis/pull/236 we found that our logic for determining if the method should be auto-paginated or not must be improved (according to the last paragraph of [AIP 4233](https://aip.dev/client-libraries/4233)).

> **Note:** If the response message has more than one non-primitive `repeated` field, the first one (in order of appearance in the file _and_ field number) is used. If the first field by order of appearance in the message and the first field by field number do not match, code generators that implement automatic pagination **should** fail with an error.

### DIFF
--- a/typescript/src/schema/proto.ts
+++ b/typescript/src/schema/proto.ts
@@ -294,30 +294,39 @@ function pagingField(messages: MessagesMap, method: MethodDescriptorProto) {
       field.label ===
       plugin.google.protobuf.FieldDescriptorProto.Label.LABEL_REPEATED
   );
-  if (repeatedFields.length !== 1) {
+  if (repeatedFields.length === 0) {
     return undefined;
+  }
+  if (repeatedFields.length === 1) {
+    return repeatedFields[0];
+  }
+  // According to https://aip.dev/client-libraries/4233, if there are two
+  // or more repeated fields in the output, we must take the the first one
+  // (in order of appearance in the file AND field number).
+  // We believe that all proto fields have numbers, hence !.
+  const minFieldNumber = repeatedFields.reduce(
+    (min: number, field: plugin.google.protobuf.IFieldDescriptorProto) => { if (field.number! < min) { min = field.number!; } return min; },
+    repeatedFields[0].number!
+  );
+  if (minFieldNumber !== repeatedFields[0].number) {
+    console.warn(`Warning: method ${method.name} has several repeated fields in the output type and violates https://aip.dev/client-libraries/4233 for auto-pagination. Disabling auto-pagination for this method.`)
+    console.warn('Fields considered for pagination:');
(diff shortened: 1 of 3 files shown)


alexander-fenster:
Also, optional chaining works! @bcoe

typescript/src/schema/proto.ts
```diff
@@ -294,30 +294,48 @@ function pagingField(messages: MessagesMap, method: MethodDescriptorProto) {
...
+    (min: number, field: plugin.google.protobuf.IFieldDescriptorProto) => {
+      if (field.number! < min) {
+        min = field.number!;
+      }
+      return min;
+    },
+    repeatedFields[0].number!
+  );
+  if (minFieldNumber !== repeatedFields[0].number) {
+    console.warn(
```
JustinBeckwith: So warnings like this always give me a moment's pause.  Chances are this will be run by a robot most of the time, and a human will often never see the output.  Realistically - do we think this merits an error, and human attention?  Or are we ok with the warning getting ignored?

typescript/src/schema/proto.ts
```diff
@@ -294,30 +294,48 @@ function pagingField(messages: MessagesMap, method: MethodDescriptorProto) {
...
 
 function pagingFieldName(messages: MessagesMap, method: MethodDescriptorProto) {
-  const repeatedFields = pagingField(messages, method);
-  if (repeatedFields && repeatedFields.name) {
-    return repeatedFields.name;
-  } else {
-    return undefined;
-  }
+  const field = pagingField(messages, method);
+  return field?.name;
```
JustinBeckwith: Good to see you living your best life here

typescript/src/schema/proto.ts
alexander-fenster: I thought about this for a while. If the API is broken in this specific way, we cannot fix it (AIPs and annotations have no way of saying "forget it, it's not auto-paginated"). AIP actually suggests to fail in this case, but then we just cannot generate this API at all, end of story.

How about this: we'll fail by default and will proceed with a command line option?

```
$ gapic-generator-typescript ... foo.proto bar.proto
Error 42: blablabla
(non-zero exit code)

$ gapic-generator-typescript --ignore-errors=42 ... foo.proto bar.proto
Warning 42: blablabla
(library generated on a best effort basis, zero exit code)
```

If you're OK with this approach, I'll first implement it (in a separate PR) and then will apply here.

JustinBeckwith: On the surface, this sounds correct.  If we know an API will be generated poorly, I'd rather us not generate at all, and harass the partner team to fix the proto.

alexander-fenster: OK! Failing now, also https://github.com/googleapis/gapic-generator-typescript/issues/158 to make it possible to ignore the error.

PR #161, opened 2019-12-05
fix(deps): pin TypeScript below 3.7.0

bcoe:


alexander-fenster:
I believe it's OK for the generator itself to stay on 3.7. I will take care of the _generated_ code in a separate PR.

PR #164, opened 2019-12-05
docs: update README

alexander-fenster:
We already had @schmidt-sebastian edit those templates, let's make it a little bit easier for folks by documenting the process a little bit.


schmidt-sebastian:
This would have been helpful to have:) Thank you for documenting! The only other issue was that my editor stripped trailing spaces on save, and hence I could not use it to edit the templates that at the time relied on trailing spaces (but that problem should be resolved in the templates, not via README).

LGTM

PR #167, opened 2019-12-09
feat: expose service stub as public member

alexander-fenster:
Fixes https://github.com/googleapis/gapic-generator/issues/2127 by exposing service stub directly. All the extra logic that needs to have a service stub can then depend on this new field.

@callmehiphop: this will help us remove changes to the generated code in Pub/Sub.


alexander-fenster:
@callmehiphop Having an exact well defined type for a stub is technically possible (since we have all the types and use them below to generate methods), but probably not in this PR.

PR #173, opened 2019-12-14
fix: use camel case for page token

schmidt-sebastian:
I believe this will fix the paging issues in Firestore's System Test.


alexander-fenster:
Thanks @schmidt-sebastian, good catch! That's a bug.

PR #176, opened 2019-12-17
feat: parse child_type resource

alexander-fenster:
This PR adds support for `.child_type`, also moving most of the resource related code into a separate `ResourceDatabase` class.

After talking to @noahdietz and @lukesneeringer, it turns out that the easiest way to support `.child_type` is to split the corresponding pattern by segments, then find all possible prefixes that correspond to known resources, and register them (to generate helper functions later).

No baseline changes (that means that the APIs included in baselines are not affected). The first known affected API is datacatalog (cc: @bcoe - this should fix it).

Note: we only support the first pattern of multi-pattern resources for now, hence `[0]` here and there. This will be taken care of separately in January, since it will be an incompatible change.

### DIFF
--- a/typescript/src/schema/proto.ts
+++ b/typescript/src/schema/proto.ts
@@ -16,6 +16,7 @@ import * as plugin from '../../../pbjs-genfiles/plugin';
 import { CommentsMap, Comment } from './comments';
 import * as objectHash from 'object-hash';
 import { milliseconds } from '../util';
+import { ResourceDescriptor, ResourceDatabase } from './resourceDatabase';
 
 const defaultNonIdempotentRetryCodesName = 'non_idempotent';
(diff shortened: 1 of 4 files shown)


typescript/src/schema/proto.ts
```diff
@@ -530,32 +523,36 @@ function augmentService(
...
+      // 1. If this resource reference has .child_type, figure out if we have any known parent resources
+      const parentResources = resourceDatabase.getParentResourcesByChildType(
+        resourceReference?.childType,
+        errorLocation
+      );
+      parentResources.map(
+        resource => (uniqueResources[resource.name] = resource)
+      );
+
+      // 2. If this resource reference has .type, we should have a know resource with this type.
```
noahdietz: Typo: s/`know`/`known`

PR #183, opened 2019-12-18
feat: set main service name

alexander-fenster:
Sometimes there are multiple services in the same API (DataCatalog is one example).

For historical reasons we need to know the "main service" to use for webpack bundle name. When we have a breaking change, I might need to change that to a reasonable autodetection, but for now, let's just specify it as a parameter if the default option is bad. 

Also, make sure services are always sorted in one particular order.

Note: I removed one baseline test because showcase API is effectively tested by a start script test and there is no need to run it twice.


alexander-fenster:
We just want to have _some_ order to prevent these changes:

image

To decide which service is the "main" (for webpack bundle) would be our job as a human (in most cases, it will be the one which name looks like the package name). We need to start using the package name for this purpose, but this change will be breaking for many libraries (e.g. translate: currently the webpack bundle is called `TranslationService`, will change to `Translate`) so we'll make it in January.

PR #184, opened 2019-12-21
fix: use camel case for path template names

alexander-fenster:
Path template names must be camel-cased, not lowercased. This issue created compatibility problem in https://github.com/googleapis/nodejs-datacatalog/issues/79.


xiaozhenliu-gg5:
Thanks Alex! does this change require any baseline test modification?

alexander-fenster:
No, we don't have any baselines that have multi-word resource names. I was thinking of adding `datacatalog` as a baseline, not sure if it's a good idea to have one more :)

PR #186, opened 2019-12-26
fix: wait for client before checking for terminate

schmidt-sebastian:
We should only check for `terminate` once the stub has been initialized. Otherwise, we might invoke the client functions if the client is terminated at startup.

Note: If I run `npm run baseline` I get a lot of unrelated changes, so I am not running this for now.


alexander-fenster:
If the unrelated baseline changes were caused by the trailing whitespace then I'm going to fix all the trailing whitespace in a separate PR. I updated baselines for you for now.

alexander-fenster:
@schmidt-sebastian Thank you! I merged it. The updated Docker image will be released on January  2nd, after holidays, and the autosynth changes will come by January 3rd.

PR #187, opened 2019-12-28
feat: use explicit mocha imports

JustinBeckwith:
With an upcoming move to `eslint`, the way we use ambient `describe` and `it` methods in `mocha` was going to require adding a `.eslintrc.yml` to every test directory, across every repository.  There's really no good reason to do this, as mocha can work just like any other npm module, and it was tripping up folks new to nodejs.  This feels more .... correct, and has the benefit of less weird eslint config. I don't know what the hell I'm doing in the generator, so let me know if I messed up a baseline or something :) 

Side note: apologies for all the trailing whitespace trimming.  I guess my editor does this by default.   I suggest hiding whitespace changes during the code review to make things cleaner.


alexander-fenster:
Thanks @JustinBeckwith! Just be prepared to see the `CHANGE ME` PRs across all TypeScript repos when we release a new Docker image after the freeze :)

JustinBeckwith:
I am submitting PRs to do it manually the best I can.  Shouldn't be too bad :)

PR #188, opened 2019-12-29
fix: trim trailing whitespace in code, templates, and generated code

alexander-fenster:
For all trailing whitespace haters (looking at you @JustinBeckwith and @schmidt-sebastian :) ), here we go :) A little bit of `find templates typescript README.md -type f | xargs perl -pi -e 's/[\s\r]+\n/\n/g'` plus a small change in the templates to make sure the generated comments also have no trailing whitespace. Enjoy!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,16 +4,16 @@
 
 This tool is a TypeScript client library generator for network APIs specified
 by [protocol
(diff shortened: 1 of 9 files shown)


schmidt-sebastian:
I don't mind trailing whitespace - I was just jealous that you got to use about half the world's trailing whitespace in one repo! :)

PR #199, opened 2020-01-14
feat: support async iterator for paging method

xiaozhenliu-gg5:
Node 10 introduces the new `async iterator` , it can be used with `for-await-of` syntax to get the response element asynchronously, on-demand for pagination method.  No breaking changes.

code example:
```
    const client = new showcase.v1beta1.EchoClient(clientOptions);
    // ... credential and auth ....
    const resources = client.pagedExpandAsync(request);
    let counter = 0;
    for await (const resource of resources){
       counter++;
       console.log(resource);
       if(counter === 10) break;
    }
```
This should wait for the change from gax: https://github.com/googleapis/gax-nodejs/pull/708
design doc: https://docs.google.com/document/d/1zKRwrHixvsSWd78rXfcs9fMuZSG3fRrd-hikE9a3P8E/edit?hl=en#


xiaozhenliu-gg5:
Client libs tests should fail because gax dos not support async iterator yet, but `text-to-sppech` API does not have paging method, so it should pass.

Hold this until PR in gax is merged googleapis/gax-nodejs#708 and loop @alexander-fenster here to track the development. Thanks

alexander-fenster:
@xiaozhenliu-gg5 You can make the generated code depend on your gax branch: 
```
"google-gax": "googleapis/gax-nodejs#page_async"
```

I actually like this idea because it will let us test it with the real generated code.  (it takes much longer to `npm install` from GitHub so might need to increase some test timeouts)

I'll mark this PR as `DO NOT MERGE` to make sure we won't accidentally merge it :)

templates/typescript_gapic/_util.njk
```diff
@@ -51,6 +59,16 @@ limitations under the License.
 {%- endfor %}
 {%- endmacro -%}
 
+
+{%- macro printPageAsyncHeader(method) %}
+ * Equivalent to {@link {{ method.name.toCamelCase() }}}, but returns an iterable object.
+ *
+ * for-await-of synctax is used with the iterable to resursively get response element on-demand.
```
alexander-fenster: `syntax`

alexander-fenster: `resursively`

templates/typescript_gapic/_util.njk
```diff
@@ -51,6 +59,16 @@ limitations under the License.
...
+ *
+ * autoPaginate option will be ignored.
```
alexander-fenster: Let's not mention this option here at all.

templates/typescript_gapic/_util.njk
```diff
@@ -163,6 +181,11 @@ limitations under the License.
  *   An object stream which emits an object representing {{ typeLink(method.pagingResponseType) }} on 'data' event.
 {%- endmacro -%}
 
+{%- macro printReturnPageAsync(method) %}
+ * @returns {Object}
+ *   An iterable Object whose iterable properties are to be iterated over.
```
alexander-fenster: This phase is cool but does not explain much to a person who has no idea :) 

Maybe write something like `An iterable Object that conforms to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols`?

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -41,6 +41,8 @@ const version = require('../../../package.json').version;
 export class {{ service.name }}Client {
   private _descriptors: Descriptors = {page: {}, stream: {}, longrunning: {}};
   private _innerApiCalls: {[name: string]: Function};
+  // tslint:disable-next-line no-any
+  private _innerCallPromise: {[name: string]: any};
```
alexander-fenster: Nooooooooooo 😨 

We need to figure out the right type here.

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -241,9 +243,9 @@ export class {{ service.name }}Client {
       '{{ method.name.toCamelCase() }}'
 {%- endfor -%}
     ];
-
+    this._innerCallPromise = {};
```
alexander-fenster: `_innerCallPromises`?

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -564,6 +566,24 @@ export class {{ service.name }}Client {
...
   }
+
+  /**
+{{- util.printCommentsPageAsync(method) }}
+ */
+
+  {{ method.name.toCamelCase() }}Async(
+      request?: {{ util.toInterface(method.inputInterface) }},
+      options?: gax.CallOptions | {}):
+    {}{
```
alexander-fenster: It should return some better type than just `{}`.

templates/typescript_gapic/_util.njk
xiaozhenliu-gg5: thanks!

xiaozhenliu-gg5:
Generated library depends on this now, so all test pass
```
"google-gax": "googleapis/gax-nodejs#page_async"
```

bcoe:
This is looking good to me, just wondering if we could potentially hold off and land it a little after the bulk of the Node 8 deprecation work, since I don't believe these changes will be breaking?

PR #200, opened 2020-01-15
DO NOT MERGE: chore: use gts 2.0

alexander-fenster:
Seems it works!

I will not merge it now because @xiaozhenliu-gg5 is working on an async iterator feature and it will be a real bummer to merge her code with these changes. We'll merge it after that PR #199 is in.

### DIFF
--- a/typescript/src/cli.ts
+++ b/typescript/src/cli.ts
@@ -14,15 +14,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import * as yargs from "yargs";
-import { Generator } from "./generator";
+import * as yargs from 'yargs';
+import {Generator} from './generator';
 
 async function main() {
   const argv = yargs.argv;
 
   if (argv.descriptor) {
-    console.error("Descriptor option is not yet supported.");
-    throw new Error("Descriptor option is not yet supported");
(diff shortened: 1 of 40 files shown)


JustinBeckwith:
Can we also remove prettier, and all the `eslint` libraries from this package.json, and the generated package.json(s)?

typescript/src/cli.ts
```diff
@@ -14,15 +14,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import * as yargs from 'yargs';
-import { Generator } from './generator';
+import * as yargs from "yargs";
```
JustinBeckwith: Gah, that isn't right :(  Can you copy the integrated prettier config from other repos into the root?  I think for the next lib, we should do `npx gts init` to let it bootstrap these files.

alexander-fenster:
Generated `package.json` is a different story, let's take care of those separately. They won't affect existing libraries anyway, only new ones, and I'll make this change when 2.0 is released (together with `engine` bump).  As for this `package.json`, neither prettier nor eslint is listed there.

typescript/src/cli.ts
alexander-fenster: Done! Looks much less scary now!

PR #207, opened 2020-01-23
DO NOT MERGE: Generate test for libraries with paging method using async iterator.

xiaozhenliu-gg5:
reference: #199 

Add test for libraries with paging method using async iterator.


xiaozhenliu-gg5:
close this PR, and merge it to interface change directly.

PR #211, opened 2020-01-24
fix: add api resource option to resourceDatabase

### LINKED ISSUE #210: Resource in the option of message is not included in path template
For irm API, annotation resource is listed in `gapic-yaml`, and defined in `incidents.proto`:
```
  - name_pattern: projects/{project}/incidents/{incident}/annotations/{annotation}
    entity_name: annotation
```
```
message Annotation {
    type: "irm.googleapis.com/Annotation"
    pattern: "projects/{project}/incidents/{incident}/annotations/{annotation}"
  };
...
}
```

but we don't catch it as a `path_template`, now we are just looking for the `resource_reference` in the field of message:
```
const resourceReference = fieldDescriptor.options?.['.google.api.resourceReference']
```
while the option of message itself should be included as well, which is in `messagesMap[property].options.['.google.api.resource']`

xiaozhenliu-gg5:
fix #210


xiaozhenliu-gg5:
Resource which is defined but not referenced causes the problem. But for `irm`, since it's shutting down, we just remove the unused resource, but if there's other use case, we will get back to this problem.

PR #212, opened 2020-01-24
fix: parameters should be camelcase in path template function

xiaozhenliu-gg5:
```
  incidentRoleAssignmentPath(
    project_id_or_number: string,
    incident_id: string,
    role_id: string
  ) {
    return this._pathTemplates.incidentroleassignmentPathTemplate.render({
      project_id_or_number,
      incident_id,
      role_id,
    });
  }
```
tslint fail for `variable name must be in lowerCamelCase or UPPER_CASE`

### DIFF
--- a/templates/typescript_gapic/src/$version/$service_client.ts.njk
+++ b/templates/typescript_gapic/src/$version/$service_client.ts.njk
@@ -587,7 +587,7 @@ export class {{ service.name }}Client {
   ) {
     return this._pathTemplates.{{ template.name.toLowerCase() }}PathTemplate.render({
 {%- for param in template.params %}
-      {{ param.toCamelCase() }}: {{ param.toCamelCase() }},
+      {{ param }}: {{ param.toCamelCase() }},
 {%- endfor %}
     });
   }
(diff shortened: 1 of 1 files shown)


alexander-fenster:
Before we do it, can you double check if `render` wants to see camel case or snake case? If `incidentroleassignmentPathTemplate` has snake case in its definition, I think we must _pass_ snake case to `render`, but the parameters of the function can be camel case. Something like this:

```
  incidentRoleAssignmentPath(
    projectIdOrNumber: string,
    incidentId: string,
    roleId: string
  ) {
    return this._pathTemplates.incidentroleassignmentPathTemplate.render({
      project_id_or_number: projectIdOrNumber,
      incident_id: incidentId,
      role_id: roleId,
    });
  }
```

alexander-fenster:
Yep that's it:

```
> const gax = require('.')
undefined

> const pt = new gax.PathTemplate('a/{b_c}/d/{e_f}')
undefined

> pt.render({'b_c': '11', 'e_f': '22'})
'a/11/d/22'
> pt.render({'bC': '11', 'eE': '22'})
Thrown:
TypeError: Value for key b_c is not provided in { bC: '11', eE: '22' }
    at /Users/fenster/nodejs-packages/gax-nodejs/build/src/pathTemplate.js:101:27
    at Array.forEach (<anonymous>)
    at PathTemplate.render (/Users/fenster/nodejs-packages/gax-nodejs/build/src/pathTemplate.js:97:23)
```
```

xiaozhenliu-gg5:
yes,  I think as your example, render needs snake_case as parameters, but for the function, it needs camelCase arguments. Making the change now, thanks!

PR #213, opened 2020-01-27
feat: support multi-pattern resources

xiaozhenliu-gg5:
Support multi_pattern resource like this, and generate proper path templates for it.
```
message InspectTemplate {
    type: "dlp.googleapis.com/InspectTemplate"
    pattern: "organizations/{organization}/inspectTemplates/{inspect_template}"
    pattern: "projects/{project}/inspectTemplates/{inspect_template}"
  };
 ...
}
```
That will generate path templates like this:
```
      organizationInspectTemplatePathTemplate: new gaxModule.PathTemplate(
        'organizations/{organization}/inspectTemplates/{inspect_template}'
      ),
      projectInspectTemplatePathTemplate: new gaxModule.PathTemplate(
        'projects/{project}/inspectTemplates/{inspect_template}'
      ),
```
**Logic:**

For resource:` {name, type, pattern: [p1, p2]}`
register resource: 
                               in type map `{type: { name, type, pattern: [p1, p2]} }  ` 
                               in pattern map `{p1: { name1, type, p1} , p2: { name2, type, p2}}`

name1, name2 for the above case is `OrganizationInspectTemplate` and `ProjectInspectTemplate`
and for searching unique resources, it looks up `type`  first in the type map, and then searches pattern map for all patterns. We do this because `resourceReference` only has `type`, no `pattern` attribute.


typescript/test/testdata/dlp/src/v2/dlp_service_client.ts.baseline
```diff
@@ -0,0 +1,3335 @@
...
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+ */
+  getJobTrigger(
+      request: protosTypes.google.privacy.dlp.v2.IGetJobTriggerRequest,
+      optionsOrCallback?: gax.CallOptions|Callback<
+          protosTypes.google.privacy.dlp.v2.IJobTrigger,
+          protosTypes.google.privacy.dlp.v2.IGetJobTriggerRequest|undefined, {}|undefined>,
+      callback?: Callback<
+          protosTypes.google.privacy.dlp.v2.IJobTrigger,
+          protosTypes.go
```
alexander-fenster: casing is wrong here (should be `organizationInspectTemplate`, not `organizationinspecttemplate`)

alexander-fenster: should be `matchInspectTemplate...`

typescript/src/schema/resourceDatabase.ts
```diff
@@ -55,32 +55,47 @@ export class ResourceDatabase {
...
+      this.patterns[patterns?.[0]] = resourceDescriptor;
+      this.types[resourceDescriptor.type!] = resourceDescriptor;
+    }
+    // resource: {name, type, pattern: [p1, p2]}
+    // register resource does: in type map {type: { name, type, pattern: [p1, p2]} }
+    //                         in pattern map {p1: { name1, type, p1} , p2: { name2, type, p2}}
+    else {
+      for (const pattern of patterns!) {
+        const params = this.getParams(pattern);
+        const name = params.map(r => r.toPascalCase()).join('');
```
alexander-fenster: I don't think we should pascal case the names here (PascalCase for resource names is kind of TypeScript specific and should go to templates). Can we leave the names in snake case (as they come in protos) and use pascal case only in templates?

typescript/src/schema/resourceDatabase.ts
```diff
@@ -155,4 +170,26 @@ export class ResourceDatabase {
 
     return result;
   }
+
+  private getParams(pattern: string): string[] {
+    const params = pattern.match(/{[a-zA-Z_]+(?:=.*?)?}/g) || [];
+    for (let i = 0; i < params.length; i++) {
```
alexander-fenster: I know that this code is just copied, but still, let's change it to `.map` :)

```ts
const params = (pattern.match(/...../) || []).map(param => param.replace(...));
```

typescript/src/schema/resourceDatabase.ts
xiaozhenliu-gg5: yes sure, so here for the pattern `organizations/{organization}/inspectTemplates/{inspect_template}` it will combine the two params into `OrganizationInspectTemplates` in pascal case, but I think `organization_inspect_template` and convert the name to pascal case in template should work. Thanks

PR #222, opened 2020-02-01
feat!: generate all api resources and resource definition which is referenced.

xiaozhenliu-gg5:
We generated all resources into pathTemplates only if:
 1. Message or fields with `google.api.resource` option
 2. Options that have `google.api.resource_definition` and are referenced by some messages or fields.

Also, sort pathTemplate to avoid random test failure.

For dlp baseline test, `organization` is not referenced in proto, so get removed by logic.
This should unblock `securitycenter` API conversion to typescript: https://github.com/googleapis/nodejs-security-center/pull/198#event-2998450223

### DIFF
--- a/typescript/src/schema/proto.ts
+++ b/typescript/src/schema/proto.ts
@@ -554,12 +554,12 @@ function augmentService(
       let resourceByType = resourceDefinitionDatabase.getResourceByType(
         resourceReference?.type
       );
-      resourceByType = resourceByType
-        ? resourceByType
-        : resourceDatabase.getResourceByType(
-            resourceReference?.type,
-            errorLocation
-          );
+      resourceByType =
+        resourceByType ??
+        resourceDatabase.getResourceByType(
+          resourceReference?.type,
+          errorLocation
+        );
       if (!resourceByType || !resourceByType.pattern) continue;
       // For multi pattern resources, we look up the type first, and get the [pattern] from resource,
       // look up pattern map for all resources.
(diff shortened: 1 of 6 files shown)


typescript/src/schema/proto.ts
```diff
@@ -526,55 +527,52 @@ function augmentService(
...
-      // 2. If this resource reference has .type, we should have a known resource with this type.
-      const resourceByType = resourceDatabase.getResourceByType(
-        resourceReference?.type,
-        errorLocation
+      // 2. If this resource reference has .type, we should have a known resource with this type, check two maps.
+      let resourceByType = resourceDefinitionDatabase.getResourceByType(
+        resourceReference?.type
       );
+      resourceByType = resourceByType
+        ? resourceByType
```
alexander-fenster: This can now be written as `??`:  

```ts
resourceByType = resourceByType ?? resourceDatabase.getResourceByType(....);
```

New language feature!

typescript/src/schema/resourceDatabase.ts
```diff
@@ -21,8 +21,8 @@ export interface ResourceDescriptor
 }
 
 export class ResourceDatabase {
-  private patterns: { [pattern: string]: ResourceDescriptor };
-  private types: { [type: string]: ResourceDescriptor };
+  patterns: { [pattern: string]: ResourceDescriptor };
```
alexander-fenster: why not private anymore?

typescript/src/schema/proto.ts
xiaozhenliu-gg5: 👍

typescript/src/schema/resourceDatabase.ts
xiaozhenliu-gg5: It's being used here in `proto.ts`:
`const allPatterns = resourceDatabase.patterns;`

PR #223, opened 2020-02-01
feat: support additional_bindings in google.api.http option

alexander-fenster:
BigQuery Storage API has this fun annotation:

```proto
  rpc CreateReadSession(CreateReadSessionRequest) returns (ReadSession) {
    option (google.api.http) = {
      post: "/v1beta1/{table_reference.project_id=projects/*}"
      body: "*"
      additional_bindings {
        post: "/v1beta1/{table_reference.dataset_id=projects/*/datasets/*}"
        body: "*"
      }
    };
    ...
  }
```

Both `table_reference.project_id` and `table_reference.dataset_id` must go into `x-goog-request-params` header, otherwise the server returns an error.

Fixing this problem. Also, the fix exposed a small problem in the page streaming test, fixing it as well.

@steffnay This, when released, will finally unblock you!

### DIFF
--- a/typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
+++ b/typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
@@ -977,7 +977,7 @@ export class KeyManagementServiceClient {
     options.otherArgs.headers[
       'x-goog-request-params'
     ] = gax.routingHeader.fromParams({
-      'crypto_key_name': request.cryptoKey!.name || '',
+      'crypto_key.name': request.cryptoKey!.name || '',
     });
     return this._innerApiCalls.updateCryptoKey(request, options, callback);
   }
@@ -1044,7 +1044,7 @@ export class KeyManagementServiceClient {
     options.otherArgs.headers[
       'x-goog-request-params'
     ] = gax.routingHeader.fromParams({
-      'crypto_key_version_name': request.cryptoKeyVersion!.name || '',
+      'crypto_key_version.name': request.cryptoKeyVersion!.name || '',
     });
     return this._innerApiCalls.updateCryptoKeyVersion(request, options, callback);
   }
--- a/typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
+++ b/typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
@@ -977,7 +977,7 @@ export class KeyManagementServiceClient {
     options.otherArgs.headers[
       'x-goog-request-params'
     ] = gax.routingHeader.fromParams({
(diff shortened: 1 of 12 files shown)


typescript/test/testdata/translate/test/gapic-translation_service-v3beta1.ts.baseline
```diff
@@ -469,8 +484,9 @@ describe('v3beta1.TranslationServiceClient', () => {
             });
             // Mock request
             const request: protosTypes.google.cloud.translation.v3beta1.IListGlossariesRequest = {};
+            request.parent = '';
             // Mock response
```
xiaozhenliu-gg5: Can I get some context of adding `request.parent` and `expectedResponse = {response: 'data'}`? It seems not related to header fix.

typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
```diff
@@ -977,7 +977,7 @@ export class KeyManagementServiceClient {
     options.otherArgs.headers[
       'x-goog-request-params'
     ] = gax.routingHeader.fromParams({
-      'crypto_key.name': request.cryptoKey!.name || '',
+      'crypto_key_name': request.cryptoKey!.name || '',
     });
```
xiaozhenliu-gg5: found this in nodejs-kms: https://github.com/googleapis/nodejs-kms/blob/master/src/v1/key_management_service_client.js#L1754
Is there a specific reason to change the params name to snakeCase?

alexander-fenster: Thank you for spotting this! It's not intended, I'll fix it.

typescript/test/testdata/translate/test/gapic-translation_service-v3beta1.ts.baseline
alexander-fenster: It partly is - the header fix triggered adding `request.parent` which made `request` not equal to `expectedResponse` and the test started failing. The fail is expected: the stream is supposed to emit a response, not a request. So I fixed the test.

xiaozhenliu-gg5: I see, thanks for clarification!

typescript/test/testdata/keymanager/src/v1/key_management_service_client.ts.baseline
xiaozhenliu-gg5: Thanks! LGTM after the fix.

alexander-fenster: Fixed now!

PR #227, opened 2020-02-05
fix: switch protoc IPATH order

summer-ji-eng:
Problem Description:
    monitoring protos are both under common library google-gax and googleapis, protoc confuse which file to pick.

    * Switch order of proto path to form a protoc command in start_script.sh

    * Add unit test for monitoring library


typescript/test/protos/google/monitoring/v3/BUILD.bazel
```diff
@@ -0,0 +1,433 @@
+# This file was automatically generated by BuildFileGenerator
```
alexander-fenster: Please `git rm` this `BUILD.bazel` file since it's not needed for running the test.

xiaozhenliu-gg5:
From test/testdata/monitoring baseline, seems the library still not generated properly. Please do not merge. Also please add a monitoring test in `circleCI` then the checks should fail, thanks!

PR #229, opened 2020-02-05
fix: remove the filter if protos in both googleapis and gax

summer-ji-eng:
remove the filter if protos in both googleapis and common library google-gax


typescript/test/unit/monitor.ts
```diff
@@ -73,7 +67,7 @@ describe('MonitoringGenerateTest', () => {
         'node ' +
           START_SCRIPT +
           ` -I${PROTOS_DIR}` +
-          ` ${MONITOR_PROTO_FILE}` +
+          ` ${MONITOR_PROTO_FILE1}` +
```
xiaozhenliu-gg5: can we just use `MONITOR_PROTO_FILE ` ?

PR #230, opened 2020-02-06
fix: throw err when fail to get service name list

summer-ji-eng:
We should throw error if generated client library is empty.

Need feedback of error message @alexander-fenster @xiaozhenliu-gg5 🙏

### DIFF
--- a/typescript/src/schema/api.ts
+++ b/typescript/src/schema/api.ts
@@ -93,6 +93,11 @@ export class API {
         this.port = port ?? this.port ?? '443';
         serviceNamesList.push(service.name || this.naming.name);
       });
+    if (serviceNamesList.length === 0) {
+      throw new Error(
+        `Can't find ${this.naming.name}'s service names, please make sure that service names are defined in the proto file.`
+      );
+    }
     this.mainServiceName = options.mainServiceName || serviceNamesList[0];
   }
 
--- a/typescript/src/schema/api.ts
+++ b/typescript/src/schema/api.ts
@@ -95,7 +95,7 @@ export class API {
       });
     if (serviceNamesList.length === 0) {
       throw new Error(
-        `Can't find ${this.naming.name}'s service names, please make sure that service names are defined in the proto file.`
+        `Can't find ${this.naming.name}'s service names, please make sure that services are defined in the proto file.`
       );
     }
     this.mainServiceName = options.mainServiceName || serviceNamesList[0];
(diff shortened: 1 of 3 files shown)


typescript/test/unit/api.ts
```diff
@@ -132,4 +140,16 @@ describe('schema/api.ts', () => {
       '../../protos/google/cloud/example/v1/test.proto',
     ]);
   });
+
+  it('should throw error when the service name is not found', () => {
+    assert.throws(() => {
```
alexander-fenster: Can you please move just the `new API()` line into `assert.throws`? This unit test checks that that exact constructor throws an exception, if anything else in this block throws for whatever reason, it should be a test failure.

```ts
it(....., { 
  stuff;
  stuff;
  stuff;
  assert.throws(() => { new API(...); });
});
```

typescript/src/schema/api.ts
```diff
@@ -93,6 +93,11 @@ export class API {
         this.port = port ?? this.port ?? '443';
         serviceNamesList.push(service.name || this.naming.name);
       });
+    if (serviceNamesList.length === 0) {
+      throw new Error(
+        `Can't find ${this.naming.name}'s service names, please make sure that service names are defined in the proto file.`
```
alexander-fenster: The error message looks good! Let's maybe say `...make sure that services are defined...`, not the service names.

xiaozhenliu-gg5:
Thanks Summer for doing this!
I have a question, since $version is also used to define the directory's name, should we check the version as well for the mainService? For example `monitoring` has an empty src folder except `index.ts` at the first time.

typescript/src/schema/api.ts
summer-ji-eng: will do

typescript/test/unit/api.ts
summer-ji-eng: agree

summer-ji-eng:
> Thanks Summer for doing this!
> I have a question, since $version is also used to define the directory's name, should we check the version as well for the mainService? For example `monitoring` has an empty src folder except `index.ts` at the first time.

If I am understanding right, I believe that $version perform as expect. Before the fix, it generated correct folder struct, like:
> . test-out-monitoring
> > ...
>> src
>>> v3
>>>> index.js
>>> index.js
...
The content is empty in the index.js.

alexander-fenster:
If there is no version, it will throw an exception in `naming.ts` being unable to parse the package name. I think we are good here!

PR #233, opened 2020-02-06
fix: include IAM into common services to ignore, refactor the code

alexander-fenster:
One more attempt to fix the common services problem properly. This time, let's move all the similar checks into one place (a public static method of the `API` class).


alexander-fenster:
@xiaozhenliu-gg5 That's right. After talking to @summer-ji-eng here (about monitoring API) I'm starting to think that it makes sense to let the generator know of any exceptions (like, "do not exclude IAM now"). We can implement it separately if we need it.

PR #234, opened 2020-02-06
fix: pass x-goog-request-params to streaming methods

alexander-fenster:
Turns out BigQuery Storage API actually cares about the `x-goog-request-params` being passed for all methods, including streaming.

Cc: @steffnay


alexander-fenster:
No special considerations, just a bug :)  All APIs tend to ignore this `x-goog-request-params`, but the BigQuery Storage API actually requires that header to be present and fails if it's not there.

PR #235, opened 2020-02-07
fix: client streaming do not have request

alexander-fenster:
This fixes the problem @hkdevandla just found with the Showcase Messaging API. The problem is not shown by our existing tests but in Messaging API we had a client streaming method that had a request parameter (which did not make any sense because client streaming methods have no request object).

@xiaozhenliu-gg5 Maybe we can add Messaging API to the Showcase tests as well?  If we add those protos to our Showcase folder it will give us better coverage almost for free.

Introduced today by me in #234 :)


alexander-fenster:
I just figured that bidi streaming APIs are also affected. Luckily, only those that have http headers set are affected (I'm guessing, that's just one call in Firestore). Anyway, I have just updated the PR to fix for bidi streaming APIs as well. @xiaozhenliu-gg5  can you take a look please?

PR #236, opened 2020-02-07
fix: copy protos in common google gax and googleapis

summer-ji-eng:
Problem: fail to copy proto files from proto.list because the duplicate in common google-gax and input proto files.

The original logic is to copy the proto files which are **not** in google-gax.

Fixed logic is to copy the proto file **either** from input proto files from cmd **or** not in google-gax, which also means that skip the proto file in gax common protos **and** not in cmd input proto files

 I didn't add extra unit test for this case, because monitoring library cover this case as well. Let me know if you have concern.
----------------------------|---------|----------|---------|---------|------------------------------
File                        | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s            
----------------------------|---------|----------|---------|---------|------------------------------              
  start_script.ts    |    93.6      |       80     |     100      |    93.6 .   | 66,67,96-98,123-125          
                   
@alexander-fenster @xiaozhenliu-gg5 @JustinBeckwith Appreciate any comment on the coding style, since I am still new to typescript. Thanks🙏

### DIFF
--- a/typescript/src/start_script.ts
+++ b/typescript/src/start_script.ts
@@ -18,7 +18,6 @@ import { execFileSync } from 'child_process';
 import * as path from 'path';
 import * as yargs from 'yargs';
 import * as fs from 'fs-extra';
-import { Z_FIXED } from 'zlib';
 const fileSystem = require('file-system');
 
 const googleGaxPath = path.dirname(require.resolve('google-gax')); // ...../google-gax/build/src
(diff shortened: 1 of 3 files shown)


typescript/src/start_script.ts
```diff
@@ -18,6 +18,7 @@ import { execFileSync } from 'child_process';
 import * as path from 'path';
 import * as yargs from 'yargs';
 import * as fs from 'fs-extra';
+import { Z_FIXED } from 'zlib';
```
alexander-fenster: this is clearly a gift from VS Code.  It sometimes inserts some random imports even before you realize that. Please remove it :)

typescript/src/start_script.ts
```diff
@@ -104,14 +105,20 @@ try {
   }
   // copy proto file to generated folder
   const protoList = path.join(outputDir, 'proto.list');
+  const protoFilesMap = new Map<string, string>(
```
alexander-fenster: I just realized you actually just need a `Set`, not a `Map`, and it can be initialized with a simple list.

PR #239, opened 2020-02-11
fix: add nested message resource to database

xiaozhenliu-gg5:
In monitoring API, we lost the resource which is the nested field of a message. Like below, we did not generate resource as `AlertPolicyCondition`

```
message AlertPolicy {
    type: "monitoring.googleapis.com/AlertPolicy"
    pattern: "projects/{project}/alertPolicies/{alert_policy}"
    pattern: "organizations/{organization}/alertPolicies/{alert_policy}"
    pattern: "folders/{folder}/alertPolicies/{alert_policy}"
    pattern: "*"
  };
 // ... other messages
  message Condition {
      type: "monitoring.googleapis.com/AlertPolicyCondition"
      pattern: "projects/{project}/alertPolicies/{alert_policy}/conditions/{condition}"
      pattern: "organizations/{organization}/alertPolicies/{alert_policy}/conditions/{condition}"
      pattern: "folders/{folder}/alertPolicies/{alert_policy}/conditions/{condition}"
      pattern: "*"
    };
   // ... other fields
  }
}
```
Now we put all messages in a stack, pop one at a time, if it has nested messages, push it into the stack until the stack is empty.


xiaozhenliu-gg5:
1. add all proto files in`monitoring/v3` and update unit test,  update baseline data.
2. add nested fields to messages stack for all resources coverage.
3. `monitoring.googleapis.com/TimeSeries` seems missing in the proto, this needs file-level resource definition.

.circleci/config.yml
```diff
@@ -139,8 +139,6 @@ jobs:
             npm run fix
             npm run compile
             npm run system-test
-            npm run docs
```
alexander-fenster: What's up with this - do they fail?

typescript/src/schema/api.ts
```diff
@@ -163,24 +163,28 @@ function getResourceDatabase(
...
         m?.options?.['.google.api.resource'] as ResourceDescriptor | undefined,
-        `file ${fd.name} message ${property}`
+        `file ${fd.name} message ${messageName}`
       );
       allResourceDatabase.registerResource(
         m?.options?.['.google.api.resource'] as ResourceDescriptor | undefined,
-        `file ${fd.name} message ${property}`
+        `file ${fd.name} message ${messageName}`
       );
+      if (m.nestedType) {
```
alexander-fenster: in one line: `(m.nestedType ?? []).map(m => messagesStack.push(m));` (not sure if it's more readable though :) )

alexander-fenster:
Thanks for catching this @xiaozhenliu-gg5!

typescript/src/schema/api.ts
xiaozhenliu-gg5: yes sure, that makes sense to me! thanks!

.circleci/config.yml
xiaozhenliu-gg5: yes, there are 4 broken links.

PR #241, opened 2020-02-12
refactor: move things around

alexander-fenster:
In this PR I'm moving a lot of things around to achieve the following goals:

* no copy-paste in baselines and `circleci.yml`
* all filenames with dashes (no underscores, no camelCase in filenames)
  - not that I'm a fan of kebab-case, I just want all files look similar :)
* `typescript/` folder contains only `*.ts` files, fixtures and protos moved out
* minor changes moving things around.


baselines/kms/proto.list.baseline
```diff
@@ -3,7 +3,7 @@ google/protobuf/descriptor.proto
 google/api/annotations.proto
 google/protobuf/duration.proto
 google/protobuf/timestamp.proto
-google/kms/v1/resources.proto
+google/cloud/kms/v1/resources.proto
```
xiaozhenliu-gg5: I'm a little bit confused here, if we generated `google/kms/v1/resources.proto` before instead of `/google/cloud/kms....` in proto_list, that sounds like a bug?

tsconfig.json
```diff
@@ -7,8 +7,5 @@
   },
   "include": [
     "typescript/**/*.ts"
-  ],
-  "exclude":[
-    "typescript/test/test_application_ts/src/*.ts"
```
xiaozhenliu-gg5: i'm not sure the reason why we excluded the file in the tsconfig here, if it's safe to remove it?

typescript/src/schema/retryable-code-map.ts
```diff
@@ -0,0 +1,137 @@
+// Copyright 2020 Google LLC
```
xiaozhenliu-gg5: This looks great, much cleaner to have a separate file for code map!

typescript/test/unit/baselines.ts
```diff
@@ -0,0 +1,74 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { describe } from 'mocha';
+import { runBaselineTest, initBaselineTest } from '../util';
+
+describe('Baseline tests', () => {
+  initBaselineTest();
+
+  runBaselineTest({
```
xiaozhenliu-gg5: This template is so cool, thanks for doing this!

.circleci/config.yml
```diff
@@ -78,15 +66,14 @@ jobs:
...
-            cd workspace/workspace/gapic-generator-typescript/.test-out-showcase
+          working_directory: /home/circleci/project/workspace/gapic-generator-typescript/.test-out-showcase
+          command: &alltests |
             npm install
             npm test
             npm run fix
             npm run compile
             npm run system-test
             npm run docs
-            npm run docs-test
```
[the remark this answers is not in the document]
alexander-fenster: Yes.  I think it does not make sense to check if links in the generated docs are valid because if they are not, it's more likely not a generator failure.  We disabled this check for DLP and might also want to disable it for monitoring, which probably means it's better to disable it for all libraries.

baselines/kms/proto.list.baseline
alexander-fenster: Originally `kms` lived in `protos/google/kms`, I just moved it to `protos/google/cloud/kms` to match the googleapis location.

tsconfig.json
alexander-fenster: Because it's now outside of `typescript` folder (it's a fixture, so it's in fixtures).

PR #280, opened 2020-02-16
chore: add all showcase protos to test for better coverage

### LINKED ISSUE #237: add Messaging API in showcase test for better coverage
refrence: https://github.com/googleapis/gapic-generator-typescript/pull/235

TODO: add the protos to our Showcase folder then it will give us better coverage almost for free.

xiaozhenliu-gg5:
fix #237 
1. add all showcase protos in fixtures.
2. remove unneeded import for `monitoring` API test.
3. modified `messaging.proto` for consistent resource type of `Blurb`.
4. update baseline dir path in `update_baseline.ts`, now it's working properly.

Warning: in `messaging.proto` resource `Blurb` has type : `google.showcase.com/Blurb` but referenced by type `showcase.googleapis.com/Blurb`.  Based on annotation rules, I changed the type to `showcase.googleapis.com/Blurb`. **But we should create a bug for it.**


xiaozhenliu-gg5:
As the description said, do you know @noahdietz where could be the best place to change the resource type in `messaging.proto`? Thank you!

noahdietz:
@xiaozhenliu-gg5 good catch! File a bug in https://github.com/googleapis/gapic-showcase/issues, or if you intend to just fix the protos, just open a PR on the repo :)

Changing the protos will result in a generated PR with some generated code changes, which I will merge, then make a release with for generators to use.

xiaozhenliu-gg5:
Sure! Thanks @noahdietz for the suggestion, will create a bug and fix the resource type there.

xiaozhenliu-gg5:
Resource type problem is resolved in `gapic-generator` in this PR https://github.com/googleapis/gapic-generator/pull/3096, so this is clean now.

xiaozhenliu-gg5:
No worries, it is not duplicate! :) @bcoe 
Thanks for the suggestion, that sounds reasonable, by `baselines/` you mean baseline data for all the generated client libraries right?
1. I think we can technically take the corresponding nodejs-$service repo as baseline and compare to it, (for example install `nodejs-dialogflow` for generating `@google-cloud/dialogflow` libraries) but I suspect that would be larger than putting all needed staff in one directory.
2. some APIs like `monitoring` which is used as baseline tests has some manual twist in `synth.py`, which may give async behaviors between the generated client and the `nodejs-monitoring` client.

but we can create a new repo for putting all baseline data into one place, not sure if it makes sense since it's only for comparison and it's part of the test coverage of the Typescript Microgenerator.

bcoe:
@xiaozhenliu-gg5 if that sounds like too much work, not a blocker for me. I was just wondering how much of this data is being pulled in from another repo, vs., is created manually.

xiaozhenliu-gg5:
in `update_basline` script, we first copy the `package.json.baseline` for updating the baseline data, and then rename it to `package.json` for renovate auto updates, finally create symlink `package.json.njk` as symlink for `package.json`.

typescript/tools/update-baselines.ts
```diff
@@ -53,12 +55,18 @@ async function copyBaseline(library: string, root: string, directory = '.') {
...
     const absolutePath = path.join(start, file);
-    const stat = await statp(absolutePath);
+    const stat = await fsstat(absolutePath);
     if (stat.isDirectory()) {
       await copyBaseline(library, root, relativePath);
     } else if (stat.isFile()) {
       const baseline = getBaselineFilename(library, relativePath);
-      await ncpp(absolutePath, baseline);
+      // In baselines/, update `package.json` instead of `package.json.baseline`
+      if (relativePath.includes('package.json')) {
```
alexander-fenster:
```
endsWith(`${path.sep}package.json`)
```
might better express what you do here

xiaozhenliu-gg5: Yes, thanks for the suggestion.

PR #281, opened 2020-02-16
fix: disable auto-pagination for Talent API

### LINKED ISSUE #279: hack: disable auto-pagination for two RPCs in Talent API
We need to hack it in the code just like here: https://github.com/googleapis/gapic-generator-go/pull/310, with the same comment, and file an issue to remove this hack later (when a new version of this API is published).

xiaozhenliu-gg5:
fix #279

### DIFF
--- a/typescript/src/schema/proto.ts
+++ b/typescript/src/schema/proto.ts
@@ -219,7 +219,7 @@ function pagingField(
   return repeatedFields[0];
 }
 
-export function pagingFieldName(
+function pagingFieldName(
   messages: MessagesMap,
   method: MethodDescriptorProto,
   service?: ServiceDescriptorProto
--- a/typescript/src/schema/proto.ts
+++ b/typescript/src/schema/proto.ts
@@ -164,7 +164,7 @@ function pagingField(
   // This should not be done for any other API.
   const serviceName =
     service && service.packageName === 'google.cloud.talent.v4beta1';
-  const methodName = method.name === 'SearchProfiles' || 'SearchJobs';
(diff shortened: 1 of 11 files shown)


typescript/src/schema/proto.ts
```diff
@@ -150,7 +150,25 @@ function streaming(method: MethodDescriptorProto) {
...
+  // TODO: remove this once the next version of the Talent API is published.
+  //
+  // This is a workaround to disable auto-pagination for specifc RPCs in
+  // Talent v4beta1. The API team will make their API non-conforming in the
+  // next version.
+  //
+  // This should not be done for any other API.
+  const serviceName =
+    service && service.packageName === 'google.cloud.talent.v4beta1';
+  const methodName = method.name === 'SearchProfiles' || 'SearchJobs';
```
alexander-fenster: This won't work! It will be parsed as `(method.name === 'SearchProfiles') || 'SearchJobs'` and will be always true, effectively skipping all the methods in Talent API.

alexander-fenster:
@xiaozhenliu-gg5 Please take your time and do not work on holiday weekends :) All these fixes can surely wait till Tuesday.

Also, just FYI, I'll probably be out on Tuesday and Wednesday, so please ask other folks (Noah and Ben) to review this when we all are back from holidays.

Thank you!

xiaozhenliu-gg5:
:) sure! Will take a look next week.

typescript/src/schema/proto.ts
xiaozhenliu-gg5: got it, changes made, and also added unit test case for better coverage.

PR #293, opened 2020-02-21
chore: install dependencies in samples/ folder for lint test

xiaozhenliu-gg5:
ref: https://github.com/googleapis/nodejs-vision/pull/532


xiaozhenliu-gg5:
This is not wanted for now, close it until we need this.

PR #296, opened 2020-02-25
feat: export protos in src/index.ts

### LINKED ISSUE #294: export protos
We need to export protos from the generated `src/index.ts`:

```ts
export * as protos from '../protos/protos';   // works in TypeScript 3.8

import * as protos from '../protos/protos';
export {protos};    // if we are still on 3.7
```

When this change is made in the generator, we'll need to make a mass PR to apply this to all libraries (including those with manual `src/index.ts` - e.g. those having multiple versions).

cc: @JustinBeckwith @bcoe

xiaozhenliu-gg5:
fix #294 
This needs further work for moving Typescript to ^3.8.0


alexander-fenster:
@sofisl @bcoe Can you folks take a look what's up with this PR? Originally, there were 2 reviewers (@summer-ji-eng and myself), I removed @summer-ji-eng then to leave just one reviewer and it still does not automerge.

PR #298, opened 2020-02-26
fix: template for configuring jsdoc

summer-ji-eng:
- correct system name using library package name
- remove source html files

### DIFF
--- a/baselines/package.json.njk
+++ b/baselines/package.json.njk
@@ -1,1 +1,0 @@
-package.json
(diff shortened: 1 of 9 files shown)


baselines/package.json.njk
```diff
@@ -0,0 +1 @@
+package.json
```
alexander-fenster: This file does not belong here (there should be nothing on this level).

PR #313, opened 2020-03-02
fix: Init mock request for server/bidi streaming method in test

xiaozhenliu-gg5:
should fix: https://github.com/googleapis/nodejs-bigquery-storage/pull/19/files
Not sure we do not have non-empty request object for server/bidi streaming method in unit test, but  this bites @google-cloud/nodejs-bigquery-storage for the first time. In our baseline test, only `showcase` has server/bidi streaming methods.


xiaozhenliu-gg5:
unit test fails because that the symlink is created for the local package.json. It seems not good for the update_baseline script. I only ran `npm run baseline` to pick up the change in template. @alexander-fenster

PR #317, opened 2020-03-05
feat: deferred initialization

### LINKED ISSUE #287: deferred client initialization
We need to move all asynchronous code (promises creation) from the client constructor to the separate `init` method that will be called by all class methods (once, at the first request). Cloud functions don't like asynchronous stuff in constructors.

To be discussed:
- if this deferred initialization should be default, optional, or be enabled only within the cloud functions
- if we want to make all our client methods `async` (possibly dropping callbacks)

alexander-fenster:
Fixes #287. 

Introduces deferred client initialization. All asynchronous code is moved away from the client constructor to a separate `initialize()` method, that will be executed when any RPC is called for the first time (subsequent `initialize()` invocations will be no-op).

Customers who want the old behavior (e.g. if they want to prepare the client and make sure the first RPC call is as fast as the subsequent ones) can call `initialize()` manually:

```ts
const client = new library.FooClient();
await client.initialize();  // optional - if not called, it will be initialized during the first call
// OR
client.initialize(); // start initializing the client, but don't wait for its completion
...
const [result] = client.rpcCall(request);
```

This is formally not a breaking change for clients, but the client behavior under the hood changes, so I recommend releasing it within the same semver major bump as Node 8 deprecation.

### DIFF
--- a/baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
+++ b/baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
@@ -37,8 +37,13 @@ export class BigQueryStorageClient {
   private _innerApiCalls: {[name: string]: Function};
   private _pathTemplates: {[name: string]: gax.PathTemplate};
   private _terminated = false;
+  private _opts: ClientOptions;
+  private _gaxModule: typeof gax | typeof gax.fallback;
+  private _gaxGrpc: gax.GrpcClient | gax.fallback.GrpcClient;
+  private _protos: {};
+  private _defaults: {[method: string]: gax.CallSettings};
   auth: gax.GoogleAuth;
-  bigQueryStorageStub: Promise<{[name: string]: Function}>;
+  bigQueryStorageStub?: Promise<{[name: string]: Function}>;
 
   /**
    * Construct an instance of BigQueryStorageClient.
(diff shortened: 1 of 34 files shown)


baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
```diff
@@ -198,6 +224,8 @@ export class BigQueryStorageClient {
         return apiCall(argument, callOptions, callback);
       };
     }
+
+    return this.bigQueryStorageStub;
```
bcoe: I assume `this._gaxModule.createApiCall`, does `initialize` not get called when you `new` a client?

baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
```diff
@@ -481,6 +511,7 @@ export class BigQueryStorageClient {
     ] = gax.routingHeader.fromParams({
       'stream.name': request.stream!.name || '',
     });
+    this.initialize();
```
bcoe: this this potentially be:

```
this.initialize().then(() => {
  return this._innerApiCalls.splitReadStream(request, options, callback);
})
```

It seems like we're still potentially creating promises that are unhandled, or maybe I'm missing something?

alexander-fenster: Yes, this does not change. We won't be able to make it perfect until we stop supporting callbacks. But now it can be more predictable since users can do `await client.initialize()` and catch all errors there.

alexander-fenster: Another problem is that the method must immediately return a _cancellable call_ that comes from Gax, not just a promise that eventually resolves to a result. 

This use case:

```
const call = client.doStuff(request);
call.cancel();  // <---- existing feature, we don't want to break it now
...
call.then(result => ...);
```

alexander-fenster:
@bcoe I suggest we go step by step here :)  This change moves out asynchronous stuff from the constructor. Bigger refactors will require gax changes related to callbacks and request cancellation, we can think about making them later but I won't be able to do it for the upcoming semver major releases :(

PR #322, opened 2020-03-11
feat: export API JSON

alexander-fenster:
**DRAFT - for design review**

A feature to dump the contents of an `API` object (will be used by samples generation).

Discussion in the shared document.

PR #326, opened 2020-03-12
refactor: allow passing multiple templates as parameters

alexander-fenster:
This is extracted from #322 to make that PR only related to samples generation.

Here I'm allowing to pass `--template` parameter pointing to a semicolon-separated list of template folders to process (by default, it's `typescript_gapic`). It will be helpful in making the generator more flexible.

### DIFF
--- a/typescript/src/start-script.ts
+++ b/typescript/src/start-script.ts
@@ -21,7 +21,10 @@ import * as fs from 'fs-extra';
 const fileSystem = require('file-system');
 
 const googleGaxPath = path.dirname(require.resolve('google-gax')); // ...../google-gax/build/src
-const GOOGLE_GAX_PROTOS_DIR = path.join(googleGaxPath, '..', '..', 'protos');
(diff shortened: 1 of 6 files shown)


typescript/src/start-script.ts
```diff
@@ -65,7 +74,7 @@ if (Array.isArray(argv._)) {
   protoFiles.push(argv._);
 }
 
-const commonProtoPath = argv.commonProtoPath || GOOGLE_GAX_PROTOS_DIR;
+const commonProtoPath = argv.commonProtoPath || googleGaxProtosDir;
```
summer-ji-eng: I curious why do we make change here. Looks like GOOGLE_GAX_PROTOS_DIR is constant variable.

alexander-fenster: We used to use `CAPITAL_LETTERS` for constants in some places but then refactored them all to regular `camelCase`, but this one was forgotten :)

PR #330, opened 2020-03-13
feat: support bundle request

xiaozhenliu-gg5:
For `logging` & `pubsub` which require bundle request, we support it from micro-generator. 
All the params are passed in GAPIC v2, batching fields will be in the structure like this:
```
interfaces: 
- name: google.logging.v2.LoggingServiceV2
  methods: 
  - name: WriteLogEntries
    batching:
      thresholds:
        element_count_threshold: 1000
        request_byte_threshold: 1048576
        delay_threshold_millis: 50
      batch_descriptor:
        batched_field: entries
        discriminator_fields:
        - log_name
        - resource
        - labels
        subresponse_field: ''// Optional
- name: name2
// ....
```


xiaozhenliu-gg5:
Copy question from design doc here: 
All params are passed as yaml through command line, except one :
```
gax.createByteLengthFunction(
protoFilesRoot.lookup('google.logging.v2.LogEntry')
)
```

LogEntry is the repeated field of the batched_field: `entries`, can I get some suggestion for the best way of getting this information `LogEntry`? is it doable to match the field `entries` in the message map and take the repeated field as this param? Thanks!

alexander-fenster:
@xiaozhenliu-gg5 I think we should take the first `repeated` field - similar to what we do for pagination.

xiaozhenliu-gg5:
Tried to copy paste the latest GAPIC v2 config to test-fixtures, and generate local logging API, unit test works well.

baselines/logging/src/v2/logging_service_v2_client.ts.baseline
```diff
@@ -0,0 +1,1799 @@
...
+          new this._gaxModule.PageDescriptor('pageToken', 'nextPageToken', 'logNames')
+    };
+    // Some methods on this API support automatically batching
+    // requests; denote this.
+    this._descriptors.batching = {
+      WriteLogEntries: new this._gaxModule.BundleDescriptor(
+        'entries',
+        ['log_name','resource','labels'],null,
+        gax.createByteLengthFunction(
+          this._opts.fallback ?
```
alexander-fenster: Just the simple `protoFilesRoot.lookup('google.logging.v2.LogEntry')` should work. Just like we do for LRO: 

```
    const waitResponse = protoFilesRoot.lookup(
      '.google.showcase.v1beta1.WaitResponse') as gax.protobuf.Type;
```

alexander-fenster: Will need to have `protoFilesRoot` defined - again, just like for LRO:

```
{%- if (service.longRunning.length > 0) %}

    // This API contains "long-running operations", which return a
    // an Operation object that allows for tracking of the operation,
    // rather than holding a request open.
    const protoFilesRoot = opts.fallback?
      this._gaxModule.protobuf.Root.fromJSON(require("../../protos/protos.json")) :
      this._gaxModule.protobuf.loadSync(nodejsProtoPath);
```

We can change the condition to work for bundling methods as well.

baselines/logging/src/v2/logging_service_v2_client_config.json.baseline
```diff
@@ -0,0 +1,46 @@
...
+        }
+      },
+      "methods": {
+        "DeleteLog": {
+          "retry_codes_name": "non_idempotent",
+          "retry_params_name": "default"
+        },
+        "WriteLogEntries": {
+          "retry_codes_name": "non_idempotent",
+          "retry_params_name": "default"
```
alexander-fenster: Bundling settings are missing here. It won't work without them.
See here: https://github.com/googleapis/nodejs-logging/blob/a3fb0f3b55583220883fb83504f94f57cf907267/src/v2/logging_service_v2_client_config.json#L42-L46

baselines/logging/src/v2/logging_service_v2_client.ts.baseline
xiaozhenliu-gg5: Trying, gax.createByteLengthFunction() takes params in this type:
```
const createByteLengthFunction: (message: {
    encode: (obj: {}) => {
        finish: () => {}[];
    };
}) => (obj: {}) => number
```

While argument of type 'Type' is not assignable to parameter of type '{ encode: (obj: {}) => { finish: () => {}[]; }; }'. It needs some type conversion I think

baselines/logging/src/v2/logging_service_v2_client_config.json.baseline
xiaozhenliu-gg5: done!

baselines/logging/src/v2/logging_service_v2_client.ts.baseline
xiaozhenliu-gg5: done.

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -218,6 +218,36 @@ export class {{ service.name }}Client {
     };
 {%- endif %}
 
+{%- if (service.bundleConfigs.length > 0) %}
+    // Some methods on this API support automatically batching
+    // requests; denote this.
+{%- if (service.longRunning.length === 0) %}
```
alexander-fenster: To reduce copy-paste, can we instead make it the part of the long running condition?

https://stackoverflow.com/questions/42677401/nunjucks-if-with-multiple-and-or-or-condition shows how to do `or` in Nunjucks.

xiaozhenliu-gg5: done! :)

PR #333, opened 2020-03-16
fix: duplicate resource name

### LINKED ISSUE #332: duplicate path Templates for resources
When testing `logging` API, the `projectPath` is added twice, which results in compile error: 
```
An object literal cannot have multiple properties with the same name in strict mode.
```
This happens in `logging_service`, `metric_service`, `config_service`

xiaozhenliu-gg5:
fix #332 
For normal resource pattern `project/{project}`, params: ['project'], resourceName: 'project', the number of elements in pattern should be double as the length of params. While for abnormal: `project/{project}/cmekSetting`, params: ['project'], just take the name from params will result in duplicate name with resource `project/{project}`. So now we append `cmekSetting` to this kind of special case.


baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -160,7 +160,7 @@ export class EchoClient {
       userPathTemplate: new this._gaxModule.PathTemplate(
         'users/{user_id}'
       ),
-      userIdBlurbIdPathTemplate: new this._gaxModule.PathTemplate(
+      userIdBlurbIdBlurbPathTemplate: new this._gaxModule.PathTemplate(
```
alexander-fenster: I understand why this one has changed, but can we make the naming logic so that it would get something like `userIdProfileBlurbId` instead? Will make more sense I guess.

alexander-fenster: The logic would be, for pairs `names/{name}`, take `name`; for parts without subsequent `{name}`, take the part name as is. So it would take `userId` for `user/{user_id}`, then `profile`, then `blurbId` for `blurbs/{blurb_id}`.

xiaozhenliu-gg5: yes, that's a side effect, I tried some ways to fix this to make it more readable but haven't get any perfect solution. I will dig deeper and try more. Thanks

xiaozhenliu-gg5: sounds reasonable to me, so for pattern `user/{user_id}/profile/blurbs/{blurb_id}`, name will be changed to `userId_profile_blurbId`.

baselines/monitoring/src/v3/metric_service_client.ts.baseline
```diff
@@ -150,7 +150,7 @@ export class MetricServiceClient {
       folderGroupPathTemplate: new this._gaxModule.PathTemplate(
         'folders/{folder}/groups/{group}'
       ),
-      folderMetricDescriptorPathTemplate: new this._gaxModule.PathTemplate(
+      folderMetricDescriptorsMetricDescriptorPathTemplate: new this._gaxModule.PathTemplate(
```
alexander-fenster: This looks wrong :) Maybe the `=**` part is confusing the logic?

xiaozhenliu-gg5: yes, match pattern does not include this case. Now it's changed back :)

baselines/monitoring/src/v3/metric_service_client.ts.baseline
```diff
@@ -1762,7 +1762,7 @@ export class MetricServiceClient {
    * Parse the metric_descriptor from OrganizationMetricDescriptor resource.
    *
    * @param {string} organizationMetricDescriptorName
-   *   A fully-qualified path representing organization_metric_descriptor resource.
+   *   A fully-qualified path representing organization_metric_descriptor=** resource.
```
alexander-fenster: Nooooooooooo

typescript/src/schema/resource-database.ts
```diff
@@ -185,6 +185,32 @@ export class ResourceDatabase {
     return params;
   }
 
+  private getName(pattern: string): string {
```
alexander-fenster: Should either be static, or just a helper function outside of the class (maybe it's better to move it to `util` and rename to `getResourceNameByPattern`?) because it does not use `this`.

typescript/src/schema/resource-database.ts
```diff
@@ -185,6 +185,32 @@ export class ResourceDatabase {
...
+    const elements = pattern.split('/');
+    const name = [];
+    // Multi pattern like: `projects/{project}/cmekSettings`, we need to append `cmekSettings` to the name.
+    // Or it will be duplicate with `project/{project}`
+    // Iterate the elements, if it comes in pairs: user/{userId}, we take `userId` as part of the name.
+    // if it comes as `profile` with no following `/{profile_id}`, we take `profile` as part of the name.
+    // So for pattern: `user/{user_id}/profile/blurbs/{blurb_id}`, name will be `userId_profile_blurbId`
+    while (elements.length > 0) {
```
alexander-fenster: Overall implementation LGTM, thank you!

baselines/monitoring/src/v3/metric_service_client.ts.baseline
xiaozhenliu-gg5: I'm on it! : ) should be quick

typescript/src/schema/resource-database.ts
xiaozhenliu-gg5: got it.

typescript/src/schema/resource-database.ts
xiaozhenliu-gg5: fixed the problem, and add `{organization=**}` case to the unit test. Thank you!

xiaozhenliu-gg5: Add one more special use case to the uni test: `{unknown_path=**}/tasks/{task}/result`. Thanks for the script of checking all odd number pairs of the resource pattern :)

PR #339, opened 2020-03-19
feat!: drop Node 8 & update Typescript to 3.8.x

xiaozhenliu-gg5:
BREAKING CHANGE: TypeScript introduced breaking changes in generated code in 3.7.x


package.json
```diff
@@ -65,12 +65,12 @@
...
-    "mocha": "^7.0.0",
+    "mocha": "^7.1.1",
     "ncp": "^2.0.0",
     "rimraf": "^3.0.2",
-    "typescript": "~3.7.5"
+    "typescript": "~3.8.3"
   },
   "engines": {
-    "node": ">=8.13.0"
+    "node": ">=10.0.0"
```
alexander-fenster: Can we do the same for all the baselines' and template `package.json` in the same PR? Drop means drop! (for the generated libraries too)

xiaozhenliu-gg5: Great suggestion! done!

xiaozhenliu-gg5: Also updated `typescript` version for the generate client library! :)

PR #340, opened 2020-03-19
feat(deps): update gts to v2

xiaozhenliu-gg5:


.prettierrc.js
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+module.exports = {
+    ...require('gts/.prettierrc.json')
+  }
```
alexander-fenster: nit, spacing

package.json
```diff
@@ -63,8 +63,12 @@
     "assert-rejects": "^1.0.0",
     "c8": "^7.1.0",
     "codecov": "^3.6.5",
+    "eslint": "^6.8.0",
+    "eslint-config-prettier": "^6.0.0",
+    "eslint-plugin-node": "^11.0.0",
+    "eslint-plugin-prettier": "^3.1.0",
     "espower-typescript": "^9.0.2",
-    "gts": "^1.1.2",
+    "gts": "^2.0.0-alpha.4",
```
alexander-fenster: why not `@next`?

alexander-fenster: Can you please try with the next? I have a feeling the formatting might change with the version change.

xiaozhenliu-gg5: oh, i ran `npm install gts@next` then it's updated like this, should I manually change it to `"gts": "@next"`? Thanks

.prettierrc.js
xiaozhenliu-gg5: done, thank you!

package.json
xiaozhenliu-gg5: updated it to `"gts": "next"`

alexander-fenster: Forget it, it's THE SAME version.

package.json
```diff
@@ -63,8 +63,12 @@
     "assert-rejects": "^1.0.0",
     "c8": "^7.1.0",
     "codecov": "^3.6.5",
+    "eslint": "^6.8.0",
```
JustinBeckwith: You can drop all of the `eslint` dependencies

typescript/src/schema/proto.ts
```diff
@@ -12,17 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/*eslint no-warning-comments: [0, { "terms": ["todo"], "location": "anywhere" }]*/
```
JustinBeckwith: Let's talk about this one folks @bcoe @alexander-fenster - do we want to broadly disable this rule?

alexander-fenster: Yes, it does not make sense for us. And having TODOs in samples is totally fine.

(this specific TODO can be removed though :) )

package.json
xiaozhenliu-gg5: removed `eslint` related dependencies.

typescript/src/schema/proto.ts
xiaozhenliu-gg5: we have `TODO: disable pagination should be removed once next version of talent API is published` in this file, I think we need this eslint here?

alexander-fenster: Oh, it's _that_ TODO. Yes, we need it.

.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "bracketSpacing": false,
```
JustinBeckwith: Shouldn't need this.  This is what the `.prettierrc.js` is supposed to do above.  If we need to make changes, we do it in `gts/.prettierrc.json` ideally.

typescript/src/cli.ts
```diff
@@ -31,6 +30,5 @@ async function main() {
 }
 
 main().catch(err => {
-  console.error(err);
-  process.exit(1);
+  throw new Error(err);
```
JustinBeckwith: Is there any value to doing this?  @bcoe   Honestly asking 😆

typescript/src/schema/proto.ts
JustinBeckwith: We yanked the upstream rule, should be able to remove this

typescript/src/start-script.ts
```diff
@@ -108,10 +108,9 @@ protocCommand.push(...protoDirsArg);
 protocCommand.push(...protoFiles);
 protocCommand.push(`-I${commonProtoPath}`);
 try {
-  execFileSync(`protoc`, protocCommand, { stdio: 'inherit' });
+  execFileSync(`protoc`, protocCommand, {stdio: 'inherit'});
 } catch (err) {
-  console.error(err.toString());
-  process.exit(1);
+  throw new Error(err.toString());
```
JustinBeckwith: No real point in catching an error just to rethrow it I'd imagine

typescript/src/start-script.ts
```diff
@@ -139,6 +138,5 @@ try {
       });
     });
 } catch (err) {
-  console.error(err.toString());
-  process.exit(1);
+  throw new Error(err.toString());
```
JustinBeckwith: Same thing as above, if we're just going to re-throw we shouldn't catch at all

typescript/src/cli.ts
alexander-fenster: That's the most confusing linter rule I ever saw :) I don't know why they hate `process.exit(1)` - do they care about non-Node environments?

typescript/src/start-script.ts
alexander-fenster: Agreed, if we don't handle the error, there is no need to catch it.

typescript/src/schema/proto.ts
xiaozhenliu-gg5: removed /*eslint no-warning-comments: .../ line since we disable this check globally in `gts`

typescript/src/cli.ts
xiaozhenliu-gg5: then can we just console.error some the err messages, instead of throwing another error? The linter hates `process.exit(1)` and suggests `throw error` for this.

.prettierrc.json
xiaozhenliu-gg5: if this line is remove, all bracket space will be brought back, but I remember we do want to remove the extra space to align with some google style? 
if that is the case, I will make the change to gts/.prettierrc.json

typescript/src/cli.ts
alexander-fenster: I guess, there is no need to catch at all. Just `main()`. The error message will be printed, together with the stack trace.

.prettierrc.json
alexander-fenster: @xiaozhenliu-gg5 I was under impression it's _already_ in gts defaults :) Is it?

xiaozhenliu-gg5: yea, I had the same impression, but if it is default as false, removing this line should not affect the linter format, right?

typescript/src/cli.ts
xiaozhenliu-gg5: got it, I will remove the unnecessary throw errors

.prettierrc.json
xiaozhenliu-gg5: just checked `gts/.prettierrc.json`, `singleQuote` is set to be true as default, but if I remove the `singleQuote: true` line, double quotes will replace single quotes. I got the feeling that if client specify a custom `.prettierrc.json` which will override the default one?

typescript/src/start-script.ts
xiaozhenliu-gg5: remove this try..catch block. Thx!

typescript/src/start-script.ts
xiaozhenliu-gg5: done. remove try..catch block.

.prettierrc.json
xiaozhenliu-gg5: removed custom `.prettierrc.json` problem solved. thank you!

typescript/src/start-script.ts
```diff
@@ -14,11 +14,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { execFileSync } from 'child_process';
+import {execFileSync} from 'child_process';
 import * as path from 'path';
 import * as yargs from 'yargs';
 import * as fs from 'fs-extra';
-const fileSystem = require('file-system');
+const fileSystem = require('file-system'); ; // eslint-disable-line
```
summer-ji-eng: nit: 
looks like we have double `；`

xiaozhenliu-gg5: ha, good catch , thank you!

PR #346, opened 2020-03-20
fix: render subresponse_field as string if it is not null

xiaozhenliu-gg5:
pubsub has subresponse_field set in batching params and it's rendering as 
```
    this._descriptors.batching = {
      publish: new gaxModule.BundleDescriptor(
        'messages',
        ['topic'],
        messageIds,
        gax.createByteLengthFunction(
          protoFilesRoot.lookup('google.pubsub.v1.PubsubMessage')
        )
      ),
    };
```
correct one should be:
```
    this._descriptors.batching = {
      publish: new gaxModule.BundleDescriptor(
        'messages',
        ['topic'],
        'messageIds',
        gax.createByteLengthFunction(
          protoFilesRoot.lookup('google.pubsub.v1.PubsubMessage')
        )
      ),
    };
```

### DIFF
--- a/templates/typescript_gapic/src/$version/$service_client.ts.njk
+++ b/templates/typescript_gapic/src/$version/$service_client.ts.njk
@@ -234,7 +234,11 @@ export class {{ service.name }}Client {
         '{{ field }}'
 {%- endfor -%}
         ],
+{%- if method.bundleConfig.batchDescriptor.subresponse_field === 'null' %}
         {{- method.bundleConfig.batchDescriptor.subresponse_field }},
+{%- else %}
+        '{{- method.bundleConfig.batchDescriptor.subresponse_field }}',
+{%- endif %}
         gax.createByteLengthFunction(
           // tslint:disable-next-line no-any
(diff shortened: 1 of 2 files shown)


templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -234,7 +234,11 @@ export class {{ service.name }}Client {
         '{{ field }}'
 {%- endfor -%}
         ],
+{%- if method.bundleConfig.batchDescriptor.subresponse_field === 'null' %}
         {{- method.bundleConfig.batchDescriptor.subresponse_field }},
```
alexander-fenster: OK so here we know it's `null`, so maybe just say `null`? :)

```
{%- if method.bundleConfig.batchDescriptor.subresponse_field === 'null' %}
        null,
```

xiaozhenliu-gg5: makes sense!

PR #343, opened 2020-03-19
feat(deps): update gts dependency for generate client library

xiaozhenliu-gg5:

PR #354, opened 2020-03-24
feat(deps): template change for new `gts` dependency

xiaozhenliu-gg5:
1. remove `PaginationCallback, PaginationResponse ` import
2. disable  any type check for `(this._protos as any)`
3. disable no-var-requires for `const serviceModule = require('../src');`
4. disable no-empty-function check for `class Operation` in gapic-test
5. remove unneeded `const expectedResponse = {};` in gapic-test

### DIFF
--- a/baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
+++ b/baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
@@ -185,8 +185,8 @@ export class BigQueryStorageClient {
     this.bigQueryStorageStub = this._gaxGrpc.createStub(
         this._opts.fallback ?
           (this._protos as protobuf.Root).lookupService('google.cloud.bigquery.storage.v1beta1.BigQueryStorage') :
-          // tslint:disable-next-line no-any
           /* eslint-disable @typescript-eslint/no-explicit-any */
+          // tslint:disable-next-line no-any
           (this._protos as any).google.cloud.bigquery.storage.v1beta1.BigQueryStorage,
         this._opts) as Promise<{[method: string]: Function}>;
 
--- a/baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
+++ b/baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
@@ -186,6 +186,7 @@ export class BigQueryStorageClient {
         this._opts.fallback ?
           (this._protos as protobuf.Root).lookupService('google.cloud.bigquery.storage.v1beta1.BigQueryStorage') :
           // tslint:disable-next-line no-any
+          /* eslint-disable @typescript-eslint/no-explicit-any */
           (this._protos as any).google.cloud.bigquery.storage.v1beta1.BigQueryStorage,
         this._opts) as Promise<{[method: string]: Function}>;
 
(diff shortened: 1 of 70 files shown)


baselines/bigquery-storage/test/gapic-big_query_storage-v1beta1.ts.baseline
```diff
@@ -19,6 +19,7 @@
 import * as protosTypes from '../protos/protos';
 import * as assert from 'assert';
 import { describe, it } from 'mocha';
+/* eslint-disable @typescript-eslint/no-var-requires */
```
alexander-fenster: Unrelated, but I would love to have a proper `import` here (I remember there were some types problems when we tried it).

baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
```diff
@@ -185,6 +185,7 @@ export class BigQueryStorageClient {
     this.bigQueryStorageStub = this._gaxGrpc.createStub(
         this._opts.fallback ?
           (this._protos as protobuf.Root).lookupService('google.cloud.bigquery.storage.v1beta1.BigQueryStorage') :
+          /* eslint-disable @typescript-eslint/no-explicit-any */
```
alexander-fenster: Do we need both `eslint-disable` and `tsline:disable`?

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -17,7 +17,7 @@
 // ** All changes to this file may be overwritten. **
 
 import * as gax from 'google-gax';
-import {APICallback, Callback, CallOptions, Descriptors, ClientOptions, LROperation, PaginationCallback, PaginationResponse} from 'google-gax';
```
alexander-fenster: Hmmmm. Maybe it's better to start actually using these types if possible? :)  (in the client template, line 551 and nearby)

baselines/bigquery-storage/test/gapic-big_query_storage-v1beta1.ts.baseline
xiaozhenliu-gg5: yes, Justin also brought this up in the other PR, import module will bring types in. Basically it's about mock methods like `client._innerApiCalls.echo`, while `_innerApiCalls` is private in `EchoClient` do we want to fix them now? Thank you!

baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
xiaozhenliu-gg5: Right! :) we just need eslint-disable, thank you! also removed tslint in base PR, thanks!

baselines/bigquery-storage/test/gapic-big_query_storage-v1beta1.ts.baseline
alexander-fenster: Can you spend an hour (but not more) on that? If we can make `innerApiCalls` public (I don't see any real harm about doing it) it might become possible to import and mock it properly. If you figure out you need more changes, then let's leave it as is.

A better alternative is to move unit tests to `sinon`, removing all the custom mocks :)  But this will take more time and we're probably not going to do it now.

baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
xiaozhenliu-gg5: it does not work without the latest version of gts. This PR was intended to make some pure template change before updating gts, but seems this is more and more overlapped with the gts PR, I may close this later on and resolve all related problems in the base PR: https://github.com/googleapis/gapic-generator-typescript/pull/343

Issue #380, opened 2020-04-02
improper sample message in unit test


xiaozhenliu-gg5:
In `nodejs-logging`, the generated unit test has improper sample message:
```
            const expectedResponse = [
              generateSampleMessage(new protosTYPE_STRING()),
              generateSampleMessage(new protosTYPE_STRING()),
              generateSampleMessage(new protosTYPE_STRING()),
            ];
```
which takes ENUM as a class to create an object, this is related to the template in util that actually checks if a type is a primitive type or a message, and substitutes the right thing.
ref: https://github.com/googleapis/nodejs-logging/pull/778

PR #382, opened 2020-04-02
fix: unit test template has wrong paging response type

xiaozhenliu-gg5:
fix: #380 
Now if the pagingResponseType is not `TYPE_MESSAGE`, we will use `.google.google.protobuf.FieldDescriptorProto.Type.TYPE_STRING (type field)` as the return type.
1.  In _util it will be truncated as `string` in primitive type check for client template. 
2. use `String` to create object for `generateSampleMessage` for unit test.

Only `logging` baseline got changed and start to use the new object.

### DIFF
--- a/baselines/logging/test/gapic_logging_service_v2_v2.ts.baseline
+++ b/baselines/logging/test/gapic_logging_service_v2_v2.ts.baseline
@@ -745,8 +745,8 @@ describe('v2.LoggingServiceV2Client', () => {
             client.descriptors.page.listLogs.createStream = stubPageStreamingCall(expectedResponse);
             const stream = client.listLogsStream(request);
             const promise = new Promise((resolve, reject) => {
-                const responses: protos.google.protobuf.FieldDescriptorProto.Type.TYPE_STRING[] = [];
-                stream.on('data', (response: protos.google.protobuf.FieldDescriptorProto.Type.TYPE_STRING) => {
+                const responses: string[] = [];
+                stream.on('data', (response: string) => {
                     responses.push(response);
                 });
                 stream.on('end', () => {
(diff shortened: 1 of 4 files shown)


templates/typescript_gapic/_util.njk
```diff
@@ -263,28 +263,41 @@ request.{{ oneComment.paramName.toCamelCase() }}
 {%- endmacro -%}
 
 {%- macro initPagingResponse(method) -%}
+{%- if method.pagingResponseType.includes('TYPE_STRING') -%}
```
alexander-fenster, line 266: Can we use `typescriptType` macro here?

baselines/logging/test/gapic_logging_service_v2_v2.ts.baseline
```diff
@@ -749,16 +741,12 @@ describe('v2.LoggingServiceV2Client', () => {
...
-              generateSampleMessage(new protosTYPE_STRING()),
-              generateSampleMessage(new protosTYPE_STRING()),
-            ];
+            const expectedResponse = [new String(), new String(), new String()];
             client.descriptors.page.listLogs.createStream = stubPageStreamingCall(expectedResponse);
             const stream = client.listLogsStream(request);
             const promise = new Promise((resolve, reject) => {
-                const responses: protosTYPE_STRING[] = [];
-                stream.on('data', (response: protosTYPE_STRING) => {
+                const responses: protos.google.protobuf.FieldDescriptorProto.Type.TYPE_STRING[] = [];
```
alexander-fenster, line 748: This does not look right - must be `string` here

templates/typescript_gapic/_util.njk
xiaozhenliu-gg5: `typescriptType` still returns type, but we need class to create new object. For `TYPE_BYTES`, return value from `typescriptType` returns `buffer`, but for `generateSampleMessage`,we need `new Buffer('')`. I was thinking about a more efficient way, do you have some suggestion here? Thanks!

baselines/logging/test/gapic_logging_service_v2_v2.ts.baseline
xiaozhenliu-gg5: fixed it!

PR #416, opened 2020-04-13
chore: update lint ignore files

JustinBeckwith:


alexander-fenster:
We don't need it here.

PR #418, opened 2020-04-14
fix: process messages from all files

alexander-fenster:
This change makes `Proto` class know about all the messages (including those from another files), not only about messages from the current file. This was triggered by Spanner backup API, specifically [ListBackups](https://github.com/googleapis/googleapis/blob/3824f547aa44df459580615c73cabb30a2a78ee0/google/spanner/admin/database/v1/spanner_database_admin.proto#L242z) RPC (the request and response messages for this call were defined in a different file and it broke the auto-pagination logic), but, as it turns out, it also adds some more resource helpers that were previously missing :)

I did not add new baseline tests, but covered the `allMessages` logic in a unit test instead.


typescript/src/schema/proto.ts
```diff
@@ -575,29 +570,15 @@ export class Proto {
   constructor(
     fd: plugin.google.protobuf.IFileDescriptorProto,
     packageName: string,
+    allMessages: MessagesMap,
     allResourceDatabase: ResourceDatabase,
     resourceDatabase: ResourceDatabase,
     options: Options
   ) {
-    fd.enumType = fd.enumType || [];
-    fd.messageType = fd.messageType || [];
```
xiaozhenliu-gg5, line 583: curious that we don't wanna handle enum type any more?

xiaozhenliu-gg5:
I have a general question, we don't touch any resource logic here, so why more resource helpers are added? (since resourceDatabase is constructed in API level)

typescript/src/schema/proto.ts
alexander-fenster: yeah, I figured, we never use enums in the generator. Less code!

alexander-fenster:
Re: general question - I don't know! :) I assume it's probably because when we generate resources, we do 

```ts
for (const property of Object.keys(messages)) {
```

and we have more `messages` to consider now (so more messages will be considered used).

typescript/src/schema/proto.ts
xiaozhenliu-gg5: sounds great! thank you!

alexander-fenster:
Oh, I see the concern. Well yes, this might be a problem. Do you think we also need to pass this file's messages for resource generation?

alexander-fenster:
@xiaozhenliu-gg5 I refactored the code a little bit, now passing both `allMessages` and `localMessages`. Also, since some functions accept several parameters with same types (two `MessageMap` and two `ResourceDatabase`, this became pretty scary so I refactored it into passing objects as parameters (makes it harder to make a mistake passing wrong thing).  Please review once more :)   No baseline changes now, and the new `allMessages` functionality is covered by a unit test.

PR #420, opened 2020-04-14
feat: decode LRO metadata & response for users

### LINKED ISSUE #170: make it possible to decode LRO response and metadata
This might be a combined issue in the generator and `google-gax`.

Users often need to get the LRO result and metadata knowing operation name, and we don't provide an API for that (it's possible but requires some manual proto decoding).

Example of this: https://github.com/googleapis/nodejs-speech/issues/399

Would be great to have a good API for LRO calls that will do this. Maybe an overload for all LROs that accept an operation name. Would be great to consult other languages and see what they do.

xiaozhenliu-gg5:
fix #170 
1. add interface for LROdecode method.
2. add comments & code samples.
3. unit test coverage.

we provide the interface for users to easily decode LROperation:
```
// name is from operation that needs to be decoded.
const decodedOperation = await decodeWaitLRO(name); 
console.log(decodedOperation.result);
console.log(decodedOperation.metadata);

```


baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -607,6 +607,11 @@ export class EchoClient {
     this.initialize();
     return this.innerApiCalls.wait(request, options, callback);
   }
+  async waitDecodeLRO(name: string){
```
alexander-fenster, line 610: Let's brainstorm on the better name with @bcoe and @JustinBeckwith. Nobody really knows what's an LRO :)

alexander-fenster:
Please also add a test for this into the test application (both TypeScript and JavaScript ones), it's the only way to verify that it works in the real life.

As for the name - let's discuss it in the chat.

Other than that - everything looks good, thank you!

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -24,7 +24,7 @@ import { Transform } from 'stream';
 import { RequestType } from 'google-gax/build/src/apitypes';
 import * as protos from '../../protos/protos';
 import * as gapicConfig from './echo_client_config.json';
-
+import * as operationProtos from 'google-gax/build/protos/operations';
```
JustinBeckwith, line 27: We should make sure whatever we need is exported from index in gax.  This pattern gets you into breaking-change trouble far too easily.

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
JustinBeckwith: Heh - what does this actually do?  Can you provide a few practical examples of what it's usage would be like, and why I'd do it?

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -607,6 +607,11 @@ export class EchoClient {
     this.initialize();
     return this.innerApiCalls.wait(request, options, callback);
   }
+  async waitDecodeLRO(name: string){
+    const [operation] = await this.operationsClient.getOperation({name}, {}) as unknown as [operationProtos.google.longrunning.Operation];
```
JustinBeckwith, line 611: All the casting happening in here makes me nervous.  what's up?

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: Sure, basically we want to create an interface for users to decode their LRO, related issue is here from speech: https://github.com/googleapis/nodejs-speech/issues/399.

Currently they have  a way to do that:
```
const [operation] = await speech.longRunningRecognize(request);
const decodedOperation = new gax.Operation(getResponse /* */, speech._descriptors.longrunning.longRunningRecognize, gax.createDefaultBackoffSettings());
console.log(decodedOperations.name)
console.log(decodedOperations.done)
console.log(decodedOperations.result)
console.log(decodedOperations.metadata)
```
But it might be too difficult for users to use some undocumented private members, so we decided to provide them a method `decodeLRO` which will do the same job. 
Micro-gen issue: https://github.com/googleapis/gapic-generator-typescript/issues/170
```
const [operation] = await speech.longRunningRecognize(request);
const decodedOperation = decodeLRO(operation.name);
```

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: yes. this can be easily exported from gax, thanks

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
alexander-fenster: I agree, this should be possible to fix (with some changes on the gax side).

@xiaozhenliu-gg5 I think the right fix here would be to upgrade `operationClient.ts` in gax to a proper TypeScript client with proper types (pretty much as you did with IAM client). Right now, `operationClient.ts` methods don't have return type declarations, so TypeScript infers the types from the method code and get much more general results than we need.

Let's make a gax PR that adds proper types to `operationsClient.ts` and then update this PR with those types.

alexander-fenster:
@JustinBeckwith TL;DR you're in a cloud function and you don't want to get billed for the whole time while you `await` for the long running operation completion. So you want to do this:

```ts
// first invocation
const [operation] = await videoIntelligenceClient.annotateVideo(cat_video);
const operationName = operation.name; // store it somewhere
```

```ts
// second invocation, in 5 minutes
const operation = await videoIntelligenceClient.XXXXX(operationName);
console.log(operaton.done);
console.log(operation.result);
```

The question is... how do we call `XXXXX`? We'll have one `XXXXX` method per each LRO method. Should it be `getAnnotateVideoOperation`, `getLongRunningRecognizeOperation`, etc.?

@xiaozhenliu-gg5 After the second look, I also think it makes perfect sense to specify the return type of this added method. It should be a generic e.g. `gax.LROperation<RequestType, MetadataType>` where `RequestType` and `MetadataType` are taken from LRO configuration.

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
JustinBeckwith: Sorry for being difficult - taking many steps back 🤣  I read through the original issue, and it starts with this:

```js
const operation = await speechClient.operationsClient.getOperation({ name });
```

I think the original problem is that this returns a serialized proto buffer.  Having such a public method directly return a serialized protobuf doesn't quite feel right to me.  Do we have this pattern in other places?

I guess my question is - why are we asking the user to jump through the decoding hoop here at all?

alexander-fenster: I explained the use case below :)  The method we're going to add here won't return any protobuf, just the real answer.

@xiaozhenliu-gg5 Maybe it's better if you provide code samples here?

xiaozhenliu-gg5: Yes sure, I will add in-code sample for this piece. Thanks!

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: Sure, I will first fix types for operationClient and then update proper type here. Thanks!

xiaozhenliu-gg5:
TODO: This depends on gax-nodejs github develop branch, change it back to latest version before merging.

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -607,6 +608,26 @@ export class EchoClient {
...
+ *   The first element of the array is an object representing the result of operations.
+ *   The second element of the array is an object representing the metadata of operations.
+ *
+ * @code sample:
+ *   const decodedOperation = await waitDecodeLRO(name);
+ *   console.log(decodedOperation.result);
+ *   console.log(decodedOperation.metadata);
+ *
+ */
+  async waitDecodeLRO(name: string): Promise<LROperation<{} | null, Metadata | null>>{
```
alexander-fenster, line 625: We still need to invent a better name. @JustinBeckwith do you still have questions about why we need this one?

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -607,6 +608,26 @@ export class EchoClient {
...
+    const request = new operationsProtos.google.longrunning.GetOperationRequest({name});
+    const [operation] = await this.operationsClient.getOperation(request);
+    const decodeOperation = new gax.Operation(operation as operationsProtos.google.longrunning.Operation, this.descriptors.longrunning.wait, gax.createDefaultBackoffSettings());
+    return decodeOperation as LROperation<{} | null, Metadata | null>;
```
alexander-fenster, line 629: Don't we know the right type? Like, for speech, the result type would be `LongRunningRecognizeResponse` and metadata would be `LongRunningRecognizeMetadata`, or something like that.

xiaozhenliu-gg5: Oh, right! I thought we wanna take the type from the definition type of `metadata & response` from `LROperation` in gax. Changing them now.

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: loop @bcoe as well for some suggestions :)

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: done.

baselines/bigquery-storage/package.json
```diff
@@ -24,7 +24,7 @@
     "test": "c8 mocha build/test"
   },
   "dependencies": {
-    "google-gax": "^2.2.0"
+    "google-gax": "git://github.com/googleapis/gax-nodejs.git#operation-client-type"
```
JustinBeckwith, line 27: Prolly want to block on that getting changed :)

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -24,7 +24,8 @@ import { Transform } from 'stream';
 import { RequestType } from 'google-gax/build/src/apitypes';
 import * as protos from '../../protos/protos';
 import * as gapicConfig from './echo_client_config.json';
-
+import { operationsProtos } from 'google-gax';
+import { Metadata } from 'google-gax/build/src/grpc';
```
JustinBeckwith, line 28: Instead, let's export this type from gax directly

baselines/disable-packing-test/test/gapic_messaging_v1beta1.ts.baseline
```diff
@@ -26,7 +26,7 @@ import * as messagingModule from '../src';
 import {PassThrough} from 'stream';
 
 import {protobuf, LROperation} from 'google-gax';
-
+import * as operationProtos from 'google-gax/build/protos/operations';
```
JustinBeckwith, line 29: Same feedback as above, let's not try to import things from nested paths

baselines/bigquery-storage/package.json
xiaozhenliu-gg5: right, I have submitted that PR this morning and will wait for the new release to unblock this. Changing the version back now.

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: Cool, I think we use more explicit types, this should be unused now, removing it.

baselines/disable-packing-test/test/gapic_messaging_v1beta1.ts.baseline
xiaozhenliu-gg5: cool! this is already exported from `gax`, so it can be imported from `google-gax` directly. Thank you!

baselines/bigquery-storage/package.json
xiaozhenliu-gg5: done.

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: we have `check*progress` and `*AwaitResult` name for now.

baselines/disable-packing-test/test/gapic_messaging_v1beta1.ts.baseline
xiaozhenliu-gg5: Waiting for the latest version of gax published.

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -607,6 +607,26 @@ export class EchoClient {
...
+ *   The first element of the array is an object representing the result of operations.
+ *   The second element of the array is an object representing the metadata of operations.
+ *
+ * @code sample:
+ *   const decodedOperation = await waitDecodeLRO(name);
+ *   console.log(decodedOperation.result);
+ *   console.log(decodedOperation.metadata);
+ *
+ */
+  async checkwaitProgress(name: string): Promise<LROperation<protos.google.showcase.v1beta1.WaitResponse, protos.google.showcase.v1beta1.WaitMetadata>>{
```
alexander-fenster, line 624: We need to `PascalCase` the method name here (I think we have a helper to do it) so that it becomes `checkWaitProgress`.

test-fixtures/test-application-ts/src/index.ts
```diff
@@ -226,3 +227,23 @@ function testWait(client: showcase.v1beta1.EchoClient) {
...
+      },
+      success: {
+        content: 'done',
+      },
+    };
+    const [operation] = await client.wait(request);
+    const decodedOperation = await client.waitDecodeLRO(operation.name!);
+    assert.deepStrictEqual(decodedOperation.name, operation.name);
+    assert(decodedOperation.metadata);
+    assert(decodedOperation.result)
```
alexander-fenster, line 247: We can actually check it to be the expected result (look at `response.content`). That way we'll also verify that the type is correct. Look at the test for `wait`.

(the same applies to the JavaScript test application)

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: you are right we have `toPascalCase()`, thanks!

test-fixtures/test-application-ts/src/index.ts
xiaozhenliu-gg5: sure, added the new checks for the content from response in test application. Thanks!

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -607,6 +607,25 @@ export class EchoClient {
     this.initialize();
     return this.innerApiCalls.wait(request, options, callback);
   }
+/**
+ * This method is for decoding operations.
```
alexander-fenster, line 611: final nit: let's change this to 
`Check the status of the long running operation returned by the wait() method. `  (note the `wait()` that is a template substitution)

xiaozhenliu-gg5: sounds cool, done! :) Thank you

baselines/bigquery-storage/package.json
```diff
@@ -24,7 +24,7 @@
     "test": "c8 mocha build/test"
   },
   "dependencies": {
-    "google-gax": "^2.2.0"
+    "google-gax": "next"
```
JustinBeckwith, line 27: Shouldn't that be a version number?

baselines/disable-packing-test/package.json
```diff
@@ -24,7 +24,7 @@
     "test": "c8 mocha build/test"
   },
   "dependencies": {
-    "google-gax": "^2.2.0"
+    "google-gax": "next"
```
JustinBeckwith, line 27: here too

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -607,6 +607,25 @@ export class EchoClient {
...
     return this.innerApiCalls.wait(request, options, callback);
   }
+/**
+ * Check the status of the long running operation returned by the wait() method.
+ * @param {String} name
+ *   The operation name that will be passed.
+ * @returns {Promise} - The promise which resolves to an object.
+ *   The decoded operation object has result and metadata field to get information from.
+ *
+ * @code sample:
```
JustinBeckwith, line 617: Tell me about this `@code` format.  How does that render in the generated `jsdoc` output?

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
```diff
@@ -607,6 +607,25 @@ export class EchoClient {
...
+ *   const decodedOperation = await checkWaitProgress(name);
+ *   console.log(decodedOperation.result);
+ *   console.log(decodedOperation.metadata);
+ *
+ */
+  async checkWaitProgress(name: string): Promise<LROperation<protos.google.showcase.v1beta1.WaitResponse, protos.google.showcase.v1beta1.WaitMetadata>>{
+    const request = new operationsProtos.google.longrunning.GetOperationRequest({name});
+    const [operation] = await this.operationsClient.getOperation(request);
+    const decodeOperation = new gax.Operation(operation as operationsProtos.google.longrunning.Operation, this.descriptors.longrunning.wait, gax.createDefaultBackoffSettings());
```
JustinBeckwith, line 626: Why the cast?

baselines/disable-packing-test/test/gapic_echo_v1beta1.ts.baseline
```diff
@@ -412,6 +412,39 @@ describe('v1beta1.EchoClient', () => {
             assert((client.innerApiCalls.wait as SinonStub)
                 .getCall(0).calledWith(request, expectedOptions, undefined));
         });
+
+        it('invokes checkWaitProgress without error', async () => {
+            const client = new echoModule.v1beta1.EchoClient({
+                credentials: {client_email: 'bogus', private_key: 'bogus'},
+                projectId: 'bogus',
+            });
+            client.initialize();
```
JustinBeckwith, line 421: Just to double check, is this sync or async?  For some reason I thought it was async.

baselines/disable-packing-test/test/gapic_echo_v1beta1.ts.baseline
```diff
@@ -412,6 +412,39 @@ describe('v1beta1.EchoClient', () => {
...
+        it('invokes checkWaitProgress with error', async () => {
+            const client = new echoModule.v1beta1.EchoClient({
+                credentials: {client_email: 'bogus', private_key: 'bogus'},
+                projectId: 'bogus',
+            });
+            client.initialize();
+            const expectedError = new Error('expected');
+
+            client.operationsClient.getOperation = stubSimpleCall(undefined, expectedError);
+            await assert.rejects(async () => { await client.checkWaitProgress(''); }, expectedError);
```
JustinBeckwith, line 444: I think this can be shortened to `await assert.rejects(client.checkWaitProgress(''), expectedError)`

JustinBeckwith: But double check me

baselines/disable-packing-test/test/gapic_echo_v1beta1.ts.baseline
```diff
@@ -26,7 +26,7 @@ import * as echoModule from '../src';
 import {PassThrough} from 'stream';
 
 import {protobuf, LROperation} from 'google-gax';
-
+import * as operationProtos from 'google-gax/build/protos/operations';
```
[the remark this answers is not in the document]
xiaozhenliu-gg5: it's wired that I can see the direct import locally but here it stays the same, let me clean this up, thank you!

baselines/bigquery-storage/package.json
xiaozhenliu-gg5: yes, I'm waiting for the new version to release. Right now it's still in `next` version

baselines/disable-packing-test/test/gapic_echo_v1beta1.ts.baseline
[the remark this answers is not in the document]
xiaozhenliu-gg5: seems this file is outdated, new changes are picked up in the new file, thanks

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: I was trying to provide a code piece, but `@example` might be a better annotation for this.

baselines/disable-packing-test/test/gapic_echo_v1beta1.ts.baseline
xiaozhenliu-gg5: you are right! tried it out and it worked perfect, thanks for the suggestion!

baselines/disable-packing-test/test/gapic_echo_v1beta1.ts.baseline
xiaozhenliu-gg5: It returns promise, yes, but we don't want to await for this, coz this will break the callback logic in the generated function calls. If we await for this, then it's sure that the calls are initialized.

alexander-fenster: To add to this - it's `async` but we are not in `async` function and cannot `await` for this. But yes, it runs in background, and all requests will internally wait for its completion.

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
alexander-fenster: Oh, yep, this cast should not be here - `this.operationsClient.getOperation()` should already return an `Operation`.

baselines/disable-packing-test/package.json
alexander-fenster: yes, we're waiting for this to become alive

baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
xiaozhenliu-gg5: so currently this `getOperation` returns `CancellablePromise<ResultTuple>`, because we wanna this promise can be cancelled.
 And the `ResultTuple` is type `[ResponseType | [ResponseType], NextPageRequestType | undefined, RawResponseType | undefined]`.

xiaozhenliu-gg5: We has an offline discussion and decided to fix this type from gax. This PR will be hold until the change is made there. Thank you!

JustinBeckwith:
Re-adding do-not-merge since this still references gax from a github uri

alexander-fenster:
LGTM! You can go ahead and merge now.

alexander-fenster:
(or have the bot do it for us :) )

alexander-fenster:
The bots have a day off today, apparently.

PR #430, opened 2020-04-18
fix: empty request params

xiaozhenliu-gg5:
fix: https://github.com/googleapis/nodejs-tasks/pull/408

In the current logic, `commentsMap` has comments in a single file, but we query the comments for a method in `allMessages` from all files. This will give us empty params if the request/return message is defined in a different file. So we construct the commentsMap in the api level and pass down it to single proto file.


typescript/src/schema/proto.ts
```diff
@@ -348,9 +348,9 @@ function augmentMethod(
       }
     }
   }
-  if (method.inputType && parameters.allMessages[method.inputType]?.field) {
+  if (method.inputType && parameters.localMessages[method.inputType]?.field) {
     const paramComment: Comment[] = [];
-    const inputType = parameters.allMessages[method.inputType!];
+    const inputType = parameters.localMessages[method.inputType!];
```
alexander-fenster, line 353: Thank you for fixing this!  I feel confused now. Was the message missing in `allMessages`? This is really weird, since `allMessages` should, by definition, contain all messages.

baselines/tasks/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "compile": "tsc -p . && cp -r protos build/",
+    "compile-protos": "compileProtos src",
+    "docs": "jsdoc -c .jsdoc.js",
+    "predocs-test": "npm run docs",
+    "docs-test": "linkinator docs",
+    "fix": "gts fix",
+    "lint": "gts check",
+    "prepare": "npm run compile-protos && npm run compile",
+    "system-test": "c8 mocha build/system-test",
+    "test": "c8 mocha build/test"
```
zamnuts, line 24: should linting be required for `test` too?
```
"test": "npm run lint && c8 mocha build/test"
```

summer-ji-eng:
Over all looks good to me. Same question Alex asked in `schema/proto.ts`

typescript/src/schema/proto.ts
xiaozhenliu-gg5: I think the empty request params in the comments are extra ones, they are added by `allMessages`, but not exist in `localMessages`

baselines/tasks/package.json
xiaozhenliu-gg5: Thanks for the suggestion! I think we have this in synthtool set up:
https://github.com/googleapis/nodejs-tasks/blob/master/synth.py#L56
also for other baseline test (nodejs-tasks API is only for testing this bug fix), we have `npm run lint` for circleCI tests: https://github.com/googleapis/gapic-generator-typescript/blob/master/.circleci/config.yml#L80 If there are not errors in `gts fix` we are good to go with `gts check` :)

PR #442, opened 2020-04-24
refactor!: bazel build

alexander-fenster:
Implementing Bazel build for gapic-generator-typescript.

You can still use `npm install`, `npm run compile`, and `npm test` (they will work with the help of [`@bazel/bazelisk`](https://www.npmjs.com/package/@bazel/bazelisk)), but having a local `bazel` installation is highly recommended.  To make `bazelisk` use the local installation, add this to your environment:

```sh
export USE_BAZEL_VERSION=`which bazel`  # path to your local Bazel
```

The code was rearranged and refactored a little bit. The main difference is that everywhere where we want to call something (`protoc`, or pass a plugin to `protoc`, or call the generator from tests), we call a shell script provided by Bazel. At the same time, I made sure that it still works if packed into Docker image with `npm pack` (with no Bazel stuff inside at all), so the transition for Yoshi should be very easy: we can keep publishing Docker images with generator updates and move libraries to Bazel build at the same time.

Please check the updated `README.md` (it shows how to call the generator using Bazel). Regular `npm install -g .` will also work, but with Bazel, you won't need to have `protoc` installed locally.

Please feel free to ask any questions :)


README.md
```diff
@@ -83,48 +83,51 @@ If you came here to make changes to the generated TypeScript libraries (e.g. `@g
...
+To compile the code:
 
 ```sh
 # in gapic-generator-typescript folder
-$ npm install  # install dependencies
-$ npm install -g .
-# make sure gapic-generator-typescript launch script in PATH
+$ npm install      # install dependencies
+$ npm run compile  # build project with Bazel
 ```
```
summer-ji-eng, line 99: I encounter $USE_BAZEL_VERSION issue. I suggest that we add line to export $USE_BAZEL_VERSION
```sh
export USE_BAZEL_VERSION=`which bazel`
```

.circleci/config.yml
```diff
@@ -2,72 +2,77 @@ version: 2.1
...
+            cp bazel-bin/test/test-application/test-*.js test-application-runners/
+            cp bazel-testlogs/unit-tests/test.outputs/outputs.zip test-application-runners/
       - persist_to_workspace:
           root: /home/circleci/project
           paths:
             - gapic-generator-typescript
             - baselines
+  lint:
+    docker:
+      - image: circleci/node:10-browsers
```
JustinBeckwith, line 39: Maybe node.js 12?

.gitignore
```diff
@@ -10,7 +10,6 @@ config.yaml
 ignored
 build
 .vscode
-package-lock.json
```
JustinBeckwith, line 13: Oh bother

BUILD.bazel
```diff
@@ -0,0 +1,166 @@
...
+    srcs = glob(["test-fixtures/**/*.*"]),
+)
+
+filegroup(
+    name = "baselines",
+    srcs = glob(["baselines/**/*.*", "baselines/**/.*"]),
+)
+
+npm_runtime_dependencies = [
+    "@npm//file-system",
```
JustinBeckwith, line 83: How does this interact with the dependencies defined in `package.json`?  Do you need to maintain this list in two places?

BUILD.bazel
```diff
@@ -0,0 +1,166 @@
...
+    "@npm//js-yaml",
+    "@npm//module-alias",
+    "@npm//nunjucks",
+    "@npm//object-hash",
+    "@npm//protobufjs",
+    "@npm//yargs",
+]
+
+npm_test_dependencies = npm_runtime_dependencies + [
+    "@npm//assert-rejects",
```
JustinBeckwith, line 96: Neither here nor there, but you should be able to drop the dependency on `assert-rejects` since `assert.rejects` is part of node core from 10.x up

BUILD.bazel
```diff
@@ -0,0 +1,166 @@
...
+    "@npm//c8",
+    "@npm//codecov",
```
JustinBeckwith, line 98: Do you specifically call into codecov, or rely on the GitHub Action like we do in libs?

BUILD.bazel
```diff
@@ -0,0 +1,166 @@
...
+    "@npm//@types/object-hash",
+    "@npm//@types/sinon",
+    "@npm//@types/yargs",
+    "@npm//gts",
+]
+
+ts_project(
+    name = "compile",
+    srcs = glob([
+        "src/**/*.ts",
```
JustinBeckwith, line 121: Is this effectively copying what's happening in `tsconfig.json`? Do we need to coordinate changes?

WORKSPACE
```diff
@@ -0,0 +1,63 @@
...
+load("@npm//:install_bazel_dependencies.bzl", "install_bazel_dependencies")
+install_bazel_dependencies()
+
+# Setup TypeScript toolchain
+load("@npm_bazel_typescript//:index.bzl", "ts_setup_workspace")
+ts_setup_workspace()
+
+load("@npm_bazel_labs//:index.bzl", "protobufjs_ts_library")
+
+yarn_install(
```
JustinBeckwith, line 55: Do you need to do explicit config to support Yarn with Bazel?

package.json
```diff
@@ -13,30 +13,27 @@
...
-    "gapic-generator-typescript": "build/src/start-script.js",
-    "protoc-gen-typescript_gapic": "build/src/cli.js"
+    "gapic-generator-typescript": "build/src/gapic-generator-typescript.js",
+    "protoc-gen-typescript_gapic": "build/src/protoc-plugin.js"
   },
   "files": [
-    "build/src",
-    "pbjs-genfiles",
-    "templates"
+    "build/**/*.js",
```
JustinBeckwith, line 20: Doesn't this mean you're shipping compiled test output?

package.json
```diff
@@ -13,30 +13,27 @@
...
+    "build/templates"
   ],
   "scripts": {
-    "baseline": "node build/tools/update-baselines.js",
-    "clean": "gts clean",
-    "codecov": "c8 --reporter=lcov mocha build/test/unit && c8 report",
-    "compile": "tsc -p .",
-    "compile-protos": "pbjs -p protos -p node_modules/google-gax/protos -t static-module -o pbjs-genfiles/plugin.js google/protobuf/compiler/plugin.proto google/api/annotations.proto google/api/field_behavior.proto google/api/resource.proto google/api/client.proto google/longrunning/operations.proto service_config.proto && pbts pbjs-genfiles/plugin.js -o pbjs-genfiles/plugin.d.ts",
+    "baseline": "node bazel-bin/tools/update-baselines.js",
+    "clean": "bazel clean && rm -rf build",
```
JustinBeckwith, line 26: I'm surprised bazel wouldn't know `build` is a build artifact and clean it up

package.json
```diff
@@ -13,30 +13,27 @@
...
+    "codecov": "c8 --reporter=lcov mocha bazel-bin/test/unit && c8 report",
+    "compile": "bazel build //...",
     "docker-test": "sh docker/test.sh",
-    "fix": "gts fix",
-    "js-test-application": "mocha build/test/test-application/test-js --timeout 600000",
-    "lint": "gts check",
-    "prepack": "cd templates/typescript_gapic && rm -f package.json.njk && mv package.json package.json.njk",
+    "fix": "gts fix src/**/*.ts test/**/*.ts tools/*.ts",
```
JustinBeckwith, line 30: Why does this need to be overidden?

package.json
```diff
@@ -45,20 +42,24 @@
     "get-stdin": "^7.0.0",
     "google-gax": "^2.3.1",
     "js-yaml": "^3.13.1",
+    "module-alias": "^2.2.2",
```
JustinBeckwith, line 45: Adding a runtime dependency in pursuit of a build toolchain change doesn't feel right

tsconfig.json
```diff
@@ -1,11 +1,21 @@
...
   "extends": "./node_modules/gts/tsconfig-google.json",
   "compilerOptions": {
-    "rootDir": "typescript",
+    "rootDir": ".",
     "outDir": "build",
-    "baseUrl": "typescript"
+    "baseUrl": ".",
+    "allowJs": true,
+    "lib": [
+      "es2016"
```
JustinBeckwith, line 9: Any reason to not just let `gts` pick `es2018`?

README.md
alexander-fenster: It's optional. It should totally work without this variable, so not mentioning it in `README`. We'll adapt to this and then can understand what's required and what's not :)

.circleci/config.yml
alexander-fenster: Well maybe yes!  (in a separate PR though - it's a separate Circle config change, this PR is already too big)

.gitignore
alexander-fenster: Yes! Because bazel. It needs reproducibility, and guess how they achieve it for Node libraries? Yes, by using a lockfile :( https://bazelbuild.github.io/rules_nodejs/#hermeticity-and-reproducibility

BUILD.bazel
alexander-fenster: Yes. https://bazelbuild.github.io/rules_nodejs/Built-ins.html
image

BUILD.bazel
alexander-fenster: Sure, since we're on Node 10, a cleanup PR is coming.

BUILD.bazel
alexander-fenster: I'm not calling it, this should be removed from here.  Actually, I have no idea how to call `codecov` and `c8` from bazel but I'll figure out :)

BUILD.bazel
alexander-fenster: TypeScript support is really wobbly in bazel :)  They use `tsconfig.json` to get some compiler options from there, but the tasks must still specifically define their inputs so that bazel could've built a dependency tree (think `make` on steroids).

WORKSPACE
alexander-fenster: So, for some reason, they [suggest](https://bazelbuild.github.io/rules_nodejs/install.html#using-bazel-managed-dependencies) using `yarn_install` over `npm_install`. The only difference is that `yarn_install` worked for me and `npm_install` did not. But I actually don't really care, as long as it downloads dependencies for us.

package.json
alexander-fenster: Hmm, it does. Let me change this.

package.json
alexander-fenster: The thing is, `build` is not a build artifact for bazel :) It puts its stuff into `bazel-*` folders that are actually symlinks pointing directly to hell.

```
$ ls -l bazel-*
lrwxr-xr-x  1 fenster  5001  131 May  8 01:58 bazel-bin -> /private/var/tmp/_bazel_fenster/a716da2979244e454b0762cc74541b5d/execroot/gapic_generator_typescript/bazel-out/darwin-fastbuild/bin
lrwxr-xr-x  1 fenster  5001  100 May  8 01:58 bazel-gapic-generator-typescript -> /private/var/tmp/_bazel_fenster/a716da2979244e454b0762cc74541b5d/execroot/gapic_generator_typescript
lrwxr-xr-x  1 fenster  5001  110 May  8 01:58 bazel-out -> /private/var/tmp/_bazel_fenster/a716da2979244e454b0762cc74541b5d/execroot/gapic_generator_typescript/bazel-out
lrwxr-xr-x  1 fenster  5001  136 May  8 01:58 bazel-testlogs -> /private/var/tmp/_bazel_fenster/a716da2979244e454b0762cc74541b5d/execroot/gapic_generator_typescript/bazel-out/darwin-fastbuild/testlogs
```

package.json
alexander-fenster: I wasn't able to make `eslint` work but now I've got an idea - maybe it just started processing `bazel-*` folders?  I'll try fixing it now.

package.json
alexander-fenster: Agreed, like 100% agreed. This is all about the `protos.js` that get compiled and placed into random location, which TypeScript can figure out during compile time but Node cannot find in the runtime. Let me add a huge TODO here, we must get rid of this, but as of today, I don't have a good idea how exactly to do it.

tsconfig.json
alexander-fenster: nope! fixed

package.json
alexander-fenster: Fixed!

package.json
alexander-fenster: Done.

BUILD.bazel
JustinBeckwith: So ... what happens here when renovate comes along and updates our `package.json` and `package-lock.json`?

alexander-fenster: Nothing bad - next time bazel build is run, it will figure out that `package-lock.json` has changed, and will fetch new dependencies. (I assume renovate knows how to deal with `package-lock.json`, I think it does)

alexander-fenster: Note that here in `BUILD` file it's just a list of packages without versions, so version updates by renovate are not going to affect this in any way.

JustinBeckwith: Ahhhh - what happens when you add a new package to `package.json`?

alexander-fenster: You go here, decide if it's a compile, or test, or runtime dependency, and add it manually to one of the lists. They call it "fine graned npm dependencies", you see.

How it works internally: after they do `yarn install` or `npm install` (whatever), they actually bazelify all the dependencies (auto-generate `BUILD.bazel` file for all of them) so that they become "targets" in bazel world (think `make` targets). Making bazel task depend on "just all the crap from `node_modules`" is not something they approve (that was deprecated long time ago), so yes, we list dependencies in two places.

BUILD.bazel
```diff
@@ -0,0 +1,164 @@
...
+    srcs = glob(["baselines/**/*.*", "baselines/**/.*"]),
+)
+
+npm_runtime_dependencies = [
+    "@npm//file-system",
+    "@npm//fs-extra",
+    "@npm//get-stdin",
+    "@npm//google-gax",
+    "@npm//js-yaml",
+    "@npm//module-alias",
```
bcoe, line 88: I find the fact that we need to declare our deps twice pretty ugly.

README.md
summer-ji-eng: Actually, `npm run compile` work without set $USE_BAZEL_VERSION on my end.

README.md
```diff
@@ -83,48 +83,51 @@ If you came here to make changes to the generated TypeScript libraries (e.g. `@g
...
-`--output_dir /tmp/translate-v3-typescript` is where to put the result  
-`--grpc-service-config google/cloud/translate/v3/translate_grpc_service_config.json`
-is an optional configuration file for timeouts and stuff  
+Line by line:
+`-I "$GOOGLEAPIS"` means pass the `googleapis` to `protoc`
+`--output-dir /tmp/translate-v3-typescript` is where to put the result
+`--grpc-service-config "$GOOGLEAPIS/google/cloud/translate/v3/translate_grpc_service_config.json"`
+is an optional configuration file for timeouts and stuff
 Then we add all the `translate` `v3` proto file to the command line, as well as the
 proto file that defines common resources (some APIs need it, some others don't).
```
summer-ji-eng, line 132: Thanks a lot updating the READ.md
Tested on my local, it works for me.
Congrats 👍

package-lock.json
```diff
@@ -0,0 +1,4835 @@
+{
```
vam-google, line 1: How is this file created and how is it supposed to be updated?

BUILD.bazel
```diff
@@ -0,0 +1,164 @@
...
+exports_files(["tsconfig.json"], visibility = ["//:__subpackages__"])
+
+package(default_visibility = ["//visibility:public"])
+
+load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary", "nodejs_test")
+load("@npm_bazel_typescript//:index.bzl", "ts_project", "ts_config")
+load("@npm_bazel_labs//:index.bzl", "protobufjs_ts_library")
+
+proto_library(
+    name = "google_protobuf_protos",
```
vam-google, line 11: The idiomatic way for bazel is to have BUIDL.bazel file in each sub-folder (which turns that folder and all its subfolders without their own BUILD.bazel files into a "package" from bazel's perspective).
For example, please consider putting proto_library targets which compile stuff under  `node_modules/google-gax/build/protos` into a BUIDL.bazel file under same directory (i.e. `node_modules/google-gax/build/protos`).

README.md
```diff
@@ -83,48 +83,51 @@ If you came here to make changes to the generated TypeScript libraries (e.g. `@g
...
 ```sh
 $ mkdir -p /tmp/translate-v3-typescript  # where to put the result
-# from googleapis folder:
-$ gapic-generator-typescript -I . \
-  --output_dir /tmp/translate-v3-typescript \
-  --grpc-service-config google/cloud/translate/v3/translate_grpc_service_config.json \
-  `find google/cloud/translate/v3 -name '*.proto'` \
-  google/cloud/common_resources.proto
+# from gapic-generator-typescript folder:
+$ bazel run //:gapic-generator-typescript -- \
```
vam-google, line 118: Please add a note that it is not the final state. Gapic-generator should not be called directly by `bazel run`, otherwise bazel looses most of its value (just becomes a way you manage gapic-generator's dependencies). The binary should be provided as an argument to `custom_proto_library` rule, similarly how it is done in Go microgenerator [go_gapic_library](https://github.com/googleapis/gapic-generator-go/blob/master/rules_go_gapic/go_gapic.bzl#L118). 

Then, all the arguments (like `translate_grpc_service_config.json` or protos themselves) become the sources of the `nodejs_gapic_library` rule itself (see [googleapis language go_gapic_library](https://github.com/googleapis/googleapis/blob/master/google/cloud/language/v1/BUILD.bazel#L112) for an example)

BUILD.bazel
```diff
@@ -0,0 +1,164 @@
...
+    extends = ["@npm//:node_modules/gts/tsconfig-google.json"],
+    source_map = True,
+    declaration = True,
+    deps = [
+        ":protos",
+    ] + npm_compile_dependencies,
+)
+
+nodejs_binary(
+    name = "protoc-plugin",
```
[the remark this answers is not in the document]
alexander-fenster: No particular reason, I can change to underscores.

package-lock.json
alexander-fenster: It's created by `npm` and it's used by bazel to ensure reproducibility. See here: https://bazelbuild.github.io/rules_nodejs/#hermeticity-and-reproducibility

BUILD.bazel
alexander-fenster: Let me postpone this as a possible refactor - I see your point but for now I would prefer not to introduce any `BUILD.bazel` files to `google-gax` (the reason for that is that `build/protos` folder in `google-gax` is auto-generated and I cannot just put stuff there; also, it's a user-facing folder).

README.md
alexander-fenster: Yes, this will be updated as soon as it works this way.

BUILD.bazel
```diff
@@ -0,0 +1,164 @@
...
+        ":compile",
+        ":templates",
+        ":protos",
+        "@com_google_protobuf//:protoc",
+    ] + npm_runtime_dependencies,
+)
+
+nodejs_test(
+    entry_point = "@npm//:node_modules/mocha/bin/mocha",
+    name = "unit-tests",
```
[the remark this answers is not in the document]
alexander-fenster: Done

BUILD.bazel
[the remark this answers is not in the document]
alexander-fenster: Done

README.md
alexander-fenster: For now, updated README.

PR #446, opened 2020-04-27
fix: disable auto-pagination for `SearchJobsForAlert` method in talent API

xiaozhenliu-gg5:
fix: #446

### DIFF
--- a/typescript/src/schema/proto.ts
+++ b/typescript/src/schema/proto.ts
@@ -164,13 +164,13 @@ function pagingField(
   // next version.
   //
   // This should not be done for any other API.
-  const serviceName =
+  const serviceNameException =
     service && service.packageName === 'google.cloud.talent.v4beta1';
-  const methodName =
+  const methodNameException =
     method.name === 'SearchProfiles' ||
     method.name === 'SearchJobs' ||
     method.name === 'SearchJobsForAlert';
-  if (serviceName && methodName) {
+  if (serviceNameException && methodNameException) {
     return undefined;
   }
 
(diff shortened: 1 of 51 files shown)


typescript/test/unit/proto.ts
```diff
@@ -113,10 +113,12 @@ describe('src/schema/proto.ts', () => {
       fd.service[0].method[1] = new plugin.google.protobuf.MethodDescriptorProto();
       fd.service[0].method[1].name = 'SearchProfiles';
       fd.service[0].method[2] = new plugin.google.protobuf.MethodDescriptorProto();
-      fd.service[0].method[2].name = 'ListJobs';
-      fd.service[0].method[2].outputType =
+      fd.service[0].method[2].name = 'SearchJobsForAlert';
```
alexander-fenster, line 116: Can we also add `SearchJobs` here?

typescript/src/schema/proto.ts
```diff
@@ -167,7 +167,9 @@ function pagingField(
   const serviceName =
```
alexander-fenster, line 167: Not directly related to this PR, but while you're here, let's maybe rename these variables to `serviceNameException` and `methodNameException`? It's weird for a variable named like `methodName` to be boolean.

typescript/test/unit/proto.ts
xiaozhenliu-gg5: sure!

typescript/src/schema/proto.ts
xiaozhenliu-gg5: yes, of course. That makes sense!

xiaozhenliu-gg5: done.

Issue #453, opened 2020-04-30
common protos are added in the generated proto list


xiaozhenliu-gg5:
The generated proto list has unneeded common protos which is in gax, but not in the generator. So including those set of protos like `google.api.annotations.prot`o will give the below error. 
```
UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open 'protos/google/api/annotations.proto'
```
It’s because that the api depends on other apis which have different paths so that the commonprefix which is used to filter the protos is incorrect.

use case: nodejs-asset, nodejs-containeranalysis.

PR #454, opened 2020-05-01
feat: add asset API to baseline test

xiaozhenliu-gg5:
1. make `runBaselineTest` accept multiple proto files.
2. add `asset` to baseline test for fixing bug.

related: #453


typescript/test/util.ts
```diff
@@ -76,10 +76,14 @@ export function initBaselineTest() {
 
 export function runBaselineTest(options: BaselineOptions) {
   const outputDir = path.join(cwd, options.outputDir);
-  const protoPath = path.join(
-    protosDirRoot,
-    options.protoPath.split('/').join(path.sep)
-  );
+  const protos = options.protoPath.split(';');
+  let protoPaths = '';
```
alexander-fenster, line 80: The intention will be a little bit more clear if you make it a list here:

`protoPaths = protos.map(proto => path.join(protosDirRoot, proto.split('/').join(path.sep)));`

and then just join it into string in where you need it in line 107.

xiaozhenliu-gg5: Great idea, making the changes now. :)

PR #455, opened 2020-05-01
fix: remove common protos in generated proto list

### LINKED ISSUE #453: common protos are added in the generated proto list

xiaozhenliu-gg5:
fix #453  This one will be merged after `asset` baseline test is added, then the diff will be easy to check.

After investigation, I noticed that for the cases that one API generation depends on other APIs which live in different directory (like `google.cloud.container-analysis` with `grafeas`, 
 `google.cloud.asset` with `google.identity`), the commonPrefix of package name will be empty or `google`. Then the common protos (`google.api.*`, `google.rpc.*`, `google.type.*`, `google.protobuf. *`) will be listed into the generated proto list (the condition check is `protoPackage.startsWith(commonPrefix)`).  This results in generated client libraries cannot find the common files. 
```
UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open 'protos/google/api/annotations.proto'
```

So in proto construction, we add one more check that if this proto file is a common proto. If it is, then we won't add them into the dependency list (since they are already in gax dependency).

This will help clean up the hacks in synth.py for `nodejs-asset` and `nodejs-containeranalysis`.


alexander-fenster:
Thank you for investigation! This makes sense.

PR #456, opened 2020-05-01
feat: fail for misconfigured service and longrunning method

### LINKED ISSUE #314: should fail for misconfigured long running operation
If an LRO RPC does not have either response type or metadata type defined, or if those types cannot be resolved, the generator should fail.

It depends on #247 (a good way to emit error and warning messages).

xiaozhenliu-gg5:
Add two checks and align the throwing error with this tool: https://github.com/googleapis/gapic-config-validator/blob/master/internal/validator/validator.go

1. One service must have a non empty `default host`;
2. Longrunning methods (which returns `.google.longrunning.Operation`) should have `operation_info` options, and the `responseType` `metadataType` should not be empty.
3. Add unit tests against new checks.
4. Run gapic-config-validator/error-conformance against the micro-gen.

fix #314


noahdietz:
> One service must have a non empty default host;

I know that some languages (Python) don't require `default_host` be present, and the AIP doesn't strictly say an error must be emitted. Idk if this is a new behavior for TS, but just wanted to make sure that was clear. FWIW all APIs are injected with it by the publish tool :)

xiaozhenliu-gg5:
@noahdietz Thanks for the insights! I also noticed that in some cases, one API does not have to have the default host, I may change it to be a warning instead of an error message. :)

typescript/src/schema/api.ts
```diff
@@ -99,7 +99,20 @@ export class API {
         servicesList.push(...fd.service!);
         return servicesList;
       }, [] as plugin.google.protobuf.IServiceDescriptorProto[])
-      .filter(service => service?.options?.['.google.api.defaultHost'])
+      .filter(service => {
+        if (!service.options || !service.options['.google.api.defaultHost']) {
+          throw Error(
```
alexander-fenster, line 104: `throw new Error` is more modern

xiaozhenliu-gg5: sure!

PR #459, opened 2020-05-04
chore: integrate config validator and error conformance testing

xiaozhenliu-gg5:
- adds the gapic-validator plugin to Docker wrapper's protoc invocation
- adds gapic-error-conformance step to CI
- tested docker image locally, if remove the default_host option for Echo Service, it will throw error like below:
```
--typescript_gapic_out: service "google.showcase.v1beta1.Echo" is missing option google.api.default_host
child_process.js:642
    throw err;
```


alexander-fenster:
@xiaozhenliu-gg5 It does not seem to work in CI, 

```
2020/05/05 16:25:36 fork/exec protoc-gen-typescript_gapic=build/src/cli.js: no such file or directory
```

can you please take a look?

xiaozhenliu-gg5:
right, it currently seems not work well with CI @alexander-fenster I will look into it. Thanks

xiaozhenliu-gg5:
@noahdietz Looking at the `-verbose` output for running `error-conformance` against `echo.proto`, even if the input file is specified as `.....echo.proto`, seems the error is about `Foo.service`. Do they have special meanings for `foo` and `bar` in this command? For typescript, proto path is set by `-I protodir absoluteProtoFile`
```
gapic-error-conformance -plugin=protoc-gen-go_gapic -plugin_opts="go-gapic-package=foo.com/bar/v1;bar"
```
If I remove the `-verbose` then output is nothing (seems successful), how can I use error conformance properly (mock foo.proto)? 
Can I get some advice here, thank you!

xiaozhenliu-gg5:
And since we are going to inject the `gapic-config-validator` in the docker image, is there a way that we can test it before merging it? Thanks @alexander-fenster

noahdietz:
>  for running error-conformance against echo.proto

`gapic-error-conformance` is compiled with it's own proto descriptors, you cannot supply a proto file to it to run against b.c there is no `protoc` invocation to compile it for each tool. Imagine `gapic-error-confirmance` **is** protoc. The goal here was to make it easy to run the given plugin and validator against a lot of controlled scenarios. The serialized `CodeGeneratorRequest` supplied to each tool contains all of the `FileDescriptor`s for each scenario. This was done because the protoc plugin interfaces says that plugins must accept a `CodeGeneratorRequest` on `stdin` and emit a `CodeGeneratorResponse` on `stdout`. Is there a difference in how the TS generator works?

The `-verbose` flag should emit the errors from each tool, the validator and the named plugin ([code](https://github.com/googleapis/gapic-config-validator/blob/master/cmd/gapic-error-conformance/main.go#L71-L72)). It looks like the TS generator plugin isn't returning any errors or isn't being run.

xiaozhenliu-gg5:
> gapic-error-conformance is compiled with its own proto descriptors, you cannot supply a proto file
> It looks like the TS generator plugin isn't returning any errors or isn't being run.
Understood, I was just trying to trigger the validator with a fake option, something like `go-gapic-package=foo.com/bar/v1;bar`. And you are right, from the stdout, TS generator is not used as plugin when running validator.

> Is there a difference in how the TS generator works?
No, same as Go, we need a CodeGeneratorRequest which should have proto files at least as input, and return CodeGeneratorResponse back.

so the minimal protoc command for showcase is (-I protoPath protoFiles)
```
protoc 
--plugin=protoc-gen-typescript_gapic=/Users/gapic-generator-typescript/build/src/cli.js
--typescript_gapic_out=/Users/gapic-generator-typescript/.test-out-showcase
-I/Users/gapic-generator-typescript/test-fixtures/protos
/Users/gapic-generator-typescript/test-fixtures/protos/google/showcase/v1beta1/echo.proto
```
And the validator command I'm using is 
```
gapic-error-conformance -plugin=protoc-gen-typescript_gapic="/Users/gapic-generator-typescript/build/src/cli.js"  -plugin_opts="-Ifoo/protos bar.proto" -verbose
```
The `cli.js` used as the real typescript plugin might be tricky here? `protoc-gen-typescript_gapic` itself is not executable, but `cli.js` is. Anything I'm doing wrong? Thank you @noahdietz

noahdietz:
How are errors emitted by `protoc-gen-typescript_gapic`? Are they returned through the [`CodeGeneratorResponse.error`](https://github.com/protocolbuffers/protobuf/blob/e667bf6eaaa2fb1ba2987c6538df81f88500d030/src/google/protobuf/compiler/plugin.proto#L100-L108) field?

`gapic-error-conformance` compares the `CodeGeneratorResponse.error` field of the generator and the validator. If `protoc-gen-typescript_gapic` _does not_ emit errors via the response, then it will never have anything to compare against.

alexander-fenster:
Oh, that makes perfect sense @noahdietz.

@xiaozhenliu-gg5 To make it emit errors properly, just wrap the main call to the generator in `try ... catch` and then if an error is caught, assign it to that response field. 

This is the great place to do it:

https://github.com/googleapis/gapic-generator-typescript/blob/master/typescript/src/generator.ts#L203-L214

noahdietz:
That was a key detail I neglected to relay to @xiaozhenliu-gg5 in our first discussion! Apologies.

xiaozhenliu-gg5:
Thanks @alexander-fenster That actually helps! 
Had an offline talk with @noahdietz and thank you for debugging with me! Now `gapic-error-conformance` is running smoothly with typescript microgenerator. 👍 The only thing is that the tool fails for packageName parsing error: issue(https://github.com/googleapis/gapic-config-validator/issues/112) we will wait for it to be merged to the tool. 

Thank you!

.circleci/config.yml
```diff
@@ -17,14 +17,24 @@ jobs:
...
+            curl -sSL https://github.com/googleapis/gapic-config-validator/releases/download/v0.5.0/gapic-config-validator-0.5.0-linux-amd64.tar.gz > config-validator.tar
+            tar xvf config-validator.tar
+            chmod +x gapic-error-conformance
+            chmod +x build/src/cli.js
+            export PATH=$PATH:`pwd`
+            gapic-error-conformance -plugin="build/src/cli.js"
       - run:
           name: Run generator unit tests
           command: |
+            echo $PATH
```
alexander-fenster, line 32: please remove this :)

.circleci/config.yml
```diff
@@ -17,14 +17,24 @@ jobs:
...
           command: |
             npm install
+      - run:
+          name: verify error conformance
+          command: |
+            curl -sSL https://github.com/googleapis/gapic-config-validator/releases/download/v0.5.0/gapic-config-validator-0.5.0-linux-amd64.tar.gz > config-validator.tar
+            tar xvf config-validator.tar
+            chmod +x gapic-error-conformance
+            chmod +x build/src/cli.js
+            export PATH=$PATH:`pwd`
```
alexander-fenster, line 27: This is very wrong. Never add `pwd` in `$PATH`. Just do `./gapic-error-conformance`.

.circleci/config.yml
```diff
@@ -17,14 +17,24 @@ jobs:
           name: Run npm install
           command: |
             npm install
+      - run:
+          name: verify error conformance
+          command: |
+            curl -sSL https://github.com/googleapis/gapic-config-validator/releases/download/v0.5.0/gapic-config-validator-0.5.0-linux-amd64.tar.gz > config-validator.tar
```
alexander-fenster, line 23: it's `.tar.gz`, not `.tar` - please fix the name here after `>` and in the next line. Also, the extract command should be `tar xzf` (with `z`). I wonder why it works without `z`?

docker/start.sh
```diff
@@ -8,12 +8,13 @@ echo "Latest commit: " 1>&2
...
-# Change directory to the input directory. 
+# Change directory to the input directory.
 # Make it easier to pass gRPC service config relative to it, e.g.
 # --grpc-service-config google/cloud/texttospeech/v1/texttospeech_grpc_service_config.json
 
-cd /in 
+cd /in
 gapic-generator-typescript \
+  --gapic-validator_out=. \
   --common-proto-path /protos/api-common-protos-master \
```
alexander-fenster, line 18: how is this option handled in `start-script.ts`?

docker/start.sh
```diff
@@ -8,12 +8,13 @@ echo "Latest commit: " 1>&2
...
 
-# Change directory to the input directory. 
+# Change directory to the input directory.
 # Make it easier to pass gRPC service config relative to it, e.g.
 # --grpc-service-config google/cloud/texttospeech/v1/texttospeech_grpc_service_config.json
 
-cd /in 
+cd /in
 gapic-generator-typescript \
+  --gapic-validator_out=. \
```
alexander-fenster, line 17: how is this option handled in `start_script.ts`?

.circleci/config.yml
xiaozhenliu-gg5: I must have `xvf` the `tar.gz`and it gives error, so I renamed to `tar.gz` to `tar` then it worked.

from online 
```
-x --extract = extract files from an archive
-v, --verbose = verbosely list files processed
-z, --gzip = gzipped files eg. for tar.gz packages
-f, --file ARCHIVE = use archive file or device ARCHIVE
```

.circleci/config.yml
xiaozhenliu-gg5: done.

.circleci/config.yml
xiaozhenliu-gg5: done.

docker/start.sh
xiaozhenliu-gg5: It will be passed to protoc command as `--gapic-validator_out=.` (added this logic)

.circleci/config.yml
```diff
@@ -1,5 +1,23 @@
 version: 2.1
 jobs:
+  validator:
```
alexander-fenster, line 3: As discussed, let's move it into the `testGenerator` task.

xiaozhenliu-gg5: 👍 doing it now, thanks

PR #461, opened 2020-05-05
feat: support non-slash resource for Ads

xiaozhenliu-gg5:
non-slash resources pattern is required to be supported across languages.

```
non_slash_resource = 'organization/{organization}/tasks/{taskId}{taskName}/result';
non_slash_resource2 = 'organization/{organization}/tasks/{taskId=*}{taskName}/result';
```


alexander-fenster:
The interesting thing here is how the path helpers should be generated. Can we somehow check it?

typescript/src/util.ts
```diff
@@ -159,12 +159,15 @@ export function getResourceNameByPattern(pattern: string): string {
...
-          )
-        );
+        const params = nextEle.match(/{[a-zA-Z_]+(?:=.*?)?}/g);
+        if (params!.length === 1) {
+          name.push(getChuckName(nextEle));
+        } else {
+          // For non-slash resource 'organization/{organization}/tasks/{task_id}{task_name}/result'
+          // Take parameters that match pattern [{task_id}, {task_name}] and combine them as part of the name.
+          const params = nextEle.match(/{[a-zA-Z_]+(?:=.*?)?}/g);
+          name.push(params?.map(p => getChuckName(p))?.join('_'));
```
alexander-fenster, line 169: We don't really need a separate function since you already have a regex here. Let's just improve your regex by putting everything _other_ than what we need into positive look-behind and look-ahead:

```
> s = "/foo/{ab}~{cd=zzz}123{efg}"
'/foo/{ab}~{cd=zzz}123{efg}'
> s.match(/(?<=\{).*?(?=(?:=.*?)?\})/g)
[ 'ab', 'cd', 'efg' ]
```

Looking closer at the regex, `/(?<=\{).*?(?=(?:=.*?)?\})`, the first opening bracket is in the positive look-behind `(?<=    \{      )`, then the part we care about follows  `.*?`  (can be changed to `[a-zA-Z_]+` if you prefer), and then a positive look-ahead eats optional trailing `=...` and the closing bracket: `(?=      (?:=.*?)?      \}    )`.

xiaozhenliu-gg5: Thank you! learned more about regex matching (I was not skilled at using ? mark)

xiaozhenliu-gg5:
Showcase test application fails because gax-nodejs does not allow non-slash resources (path template parser), I'm rewriting the logic for it here: https://github.com/googleapis/gax-nodejs/pull/823
The support for non-slash resources will be added after that, and now this PR is blocked.

xiaozhenliu-gg5:
https://github.com/googleapis/gax-nodejs/pull/833

PR #523, opened 2020-05-27
fix: docker test

xiaozhenliu-gg5:


src/gapic-generator-typescript.ts
```diff
@@ -101,7 +101,7 @@ const protocCommand = [
   `--plugin=protoc-gen-typescript_gapic=${protocPlugin}`,
   `--typescript_gapic_out=${outputDir}`,
 ];
-if (gapicValidatorOut && validation !== 'true') {
+if (gapicValidatorOut && validation === 'true') {
```
alexander-fenster, line 104: oopsie - I missed this in the previous PR :)

xiaozhenliu-gg5: :) even missed by myself that the option should be flipped. Now the docker test script works!

summer-ji-eng: I was so confused when I take look at https://github.com/googleapis/gapic-generator-typescript/pull/522/files
💯

PR #524, opened 2020-05-27
fix: fix the default value of validation option

xiaozhenliu-gg5:
TWIL: `const validation = argv.validation as string | 'true';` is not setting the default value to be true.
`const validation = argv.validation ?? 'true';` does the job.

### DIFF
--- a/src/gapic-generator-typescript.ts
+++ b/src/gapic-generator-typescript.ts
@@ -80,7 +80,7 @@ const packageName = argv.packageName as string | undefined;
 const mainServiceName = argv.mainService as string | undefined;
 const template = argv.template as string | undefined;
 const gapicValidatorOut = argv.gapicValidatorOut as string | undefined;
-const validation = argv.validation as string | undefined;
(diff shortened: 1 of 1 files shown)


src/gapic-generator-typescript.ts
```diff
@@ -80,7 +80,7 @@ const packageName = argv.packageName as string | undefined;
 const mainServiceName = argv.mainService as string | undefined;
 const template = argv.template as string | undefined;
 const gapicValidatorOut = argv.gapicValidatorOut as string | undefined;
-const validation = argv.validation as string | 'true';
+const validation = argv.validation as string | undefined;
```
alexander-fenster, line 83: The modern way to do it is `const validation = (argv.validation as string | undefined) ?? 'true'`, then in the condition just check `... && validation === 'true'` and it will work as expected.

xiaozhenliu-gg5: Thank you for the advice, done.

PR #535, opened 2020-06-03
feat: update keywords in package.json

### LINKED ISSUE #114: populate keywords in package.json
our old generator made an effort to populate keywords in the generated package.json.

xiaozhenliu-gg5:
fix #114

### DIFF
--- a/src/schema/api.ts
+++ b/src/schema/api.ts
@@ -33,6 +33,7 @@ export class API {
   // For historical reasons, Webpack library name matches "the main" service of the client library.
   // Sometimes it's hard to figure out automatically, so making this an option.
   mainServiceName: string;
+  serviceNamesList: string[];
 
   static isIgnoredService(
     fd: protos.google.protobuf.IFileDescriptorProto
@@ -91,7 +92,7 @@ export class API {
         return map;
       }, {} as ProtosMap);
 
-    const serviceNamesList: string[] = [];
+    this.serviceNamesList = [];
     fileDescriptors
       .filter(fd => !API.isIgnoredService(fd))
       .filter(fd => fd.service)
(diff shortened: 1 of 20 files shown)


xiaozhenliu-gg5:
Related issue can be found [here](https://github.com/eslint/eslint/issues/10364): that malformed packags.json(expected) will fail the eslint check.

templates/typescript_gapic/package.json
```diff
@@ -10,6 +10,23 @@
...
+    "google api client",
+    "google apis",
+    "google api",
+    "google",
+    "google cloud platform",
+    "google cloud",
+    "cloud",
+    "google {{ api.naming.productName.toCamelCase() }}",
+    "{{ api.naming.productName.toCamelCase() }}",
+{%- set lastItem = api.uniqkeywords.pop() -%}
```
alexander-fenster, line 24: This is actually not needed. Remove the comma in line 23, and in your template, put the comma _before_ the `{{ words }}`, not after. It will work. Don't care about formatting since we always post-format JSONs after generation.

src/schema/api.ts
```diff
@@ -132,6 +133,10 @@ export class API {
       );
     }
     this.mainServiceName = options.mainServiceName || serviceNamesList[0];
+    // For generating keywords in package.json
+    this.uniqkeywords = [
```
alexander-fenster, line 137: nit: this should be `uniqKeywords`, we use camel case for variables everywhere

xiaozhenliu-gg5: 👍

templates/typescript_gapic/package.json
xiaozhenliu-gg5: Thank you! I was struggling with the format and line splitter.

PR #543, opened 2020-06-04
chore: add documentation for bundling config

xiaozhenliu-gg5:
Users might be confused about how to set bundling config from client side.
Related issues:
1) https://github.com/googleapis/gax-nodejs/issues/160
2) https://github.com/googleapis/gax-nodejs/issues/256
So we added some documentations for overwriting the bundling options in template which users can use as a reference.


baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -84,6 +84,13 @@ export class AssetServiceClient {
     }
     opts.servicePath = opts.servicePath || servicePath;
     opts.port = opts.port || port;
+
+    // users can override the config from client side, like bundling options.
+    // The detailed structure of the clientConfig can be found here: https://github.com/googleapis/gax-nodejs/blob/master/src/gax.ts#L546
+    // The way to override bundling options:
+    //
+    // const customConfig = {"interfaces": {"service": {"methods": {"methodName": {"bundling": {..}}}}}}
+    // const client = new Client({ projectId, customConfig });
```
alexander-fenster, line 93: we can use the real client name here instead of `Client`. Also, let's not talk about `bundling` because it's a very special use case (only for logging and pubsub). Let's just say that the config can be overridden.

baselines/disable-packing-test/protos/google/showcase/v1beta1/identity.proto.baseline
```diff
@@ -102,24 +102,23 @@ message User {
     (google.api.field_behavior) = OUTPUT_ONLY
   ];
 
-  // Typescript generator doe not support proto3 optional fields
   // The age of the use in years.
-  // optional int32 age = 6;
+  optional int32 age = 6;
```
alexander-fenster, line 106: This is interesting :)  I forgot to update this baseline proto file in #539 but the tests did not fail. Do we ignore proto files when we compare baselines?

baselines/asset/src/v1/asset_service_client.ts.baseline
alexander-fenster: We can even use a real service name and a real method name (e.g. the first one).

baselines/disable-packing-test/protos/google/showcase/v1beta1/identity.proto.baseline
xiaozhenliu-gg5: I think you updated this file in that PR, at least from here:
https://github.com/googleapis/gapic-generator-typescript/pull/539/files#diff-755cd9dd88b3024c1452c8792a91db45L107

alexander-fenster: No, I updated `baselines/showcase` but not `baselines/disable-packing-test`. It should have failed but it did not :)  This is unrelated to this PR, but seems like the baseline test does not always detect proto files changes for whatever reason.

baselines/asset/src/v1/asset_service_client.ts.baseline
xiaozhenliu-gg5: Ok, so we wanna give example of other config with the real name, maybe `showcase`

baselines/disable-packing-test/protos/google/showcase/v1beta1/identity.proto.baseline
xiaozhenliu-gg5: ummmm, I see, that's also updated from another PR, when I run `npm run baseline`, it updates the library.proto.baseline.

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -84,6 +84,13 @@ export class AssetServiceClient {
     }
     opts.servicePath = opts.servicePath || servicePath;
     opts.port = opts.port || port;
+
+    // users can override the config from client side, like retry codes name, bundling options.
+    // The detailed structure of the clientConfig can be found here: https://github.com/googleapis/gax-nodejs/blob/master/src/gax.ts#L546
+    // The way to override client config for Showcase API:
+    //
+    // const customConfig = {"interfaces": {"google.showcase.v1beta1.Echo": {"methods": {"Echo": {"retry_codes_name": "idempotent", "retry_params_name": "default", "bundling": {..}}}}}}
```
alexander-fenster, line 92: the final nit, let's completely remove `bundling` from the example, it will only confuse users :)

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -84,6 +84,13 @@ export class AssetServiceClient {
...
+    // const showcaseClient = showcaseClient({ projectId, customConfig });
```
alexander-fenster, line 93: `= new ShowcaseClient`

baselines/asset/src/v1/asset_service_client.ts.baseline
xiaozhenliu-gg5: cool, thanks!

baselines/asset/src/v1/asset_service_client.ts.baseline
xiaozhenliu-gg5: good catch, thank you!

PR #565, opened 2020-06-11
fix: make fallback work properly in Electron

alexander-fenster:
This will properly fix the problem reported in https://github.com/googleapis/nodejs-speech/issues/547 and https://github.com/googleapis/nodejs-vision/issues/573. The existing code did not work for Electron (which has `window` defined but does not follow `browser` field in `package.json` so it picked the wrong non-fallback `google-gax` implementation).

The new code is just simpler. It should work properly for webpack+browsers (since `webpack` respects `browser` field in `google-gax` and will get the fallback implementation), and for Electron, users will need to pass `{fallback: true}` to the client constructor.

### DIFF
--- a/baselines/asset/src/v1/asset_service_client.ts.baseline
+++ b/baselines/asset/src/v1/asset_service_client.ts.baseline
@@ -93,14 +93,11 @@ export class AssetServiceClient {
     // const showcaseClient = new showcaseClient({ projectId, customConfig });
     opts.clientConfig = opts.clientConfig || {};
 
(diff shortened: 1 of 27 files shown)


bcoe:
@alexander-fenster any chance we could do a mass PR for this, the fact that changes come in form the generator with no context is really starting to be a pain for daily updates. It's much more noticeable now that we have context for googleapis/synthtool, and the googleapis/googleapis repos.

alexander-fenster:
@bcoe Yes, I'll make a mass PR, no problem!

PR #574, opened 2020-06-24
fix: make plugin parameters work properly

alexander-fenster:
Handling extra plugin parameters properly.

Also, adding a debug print to simplify transitioning from Docker containers to Bazel builds.

### DIFF
--- a/src/generator.ts
+++ b/src/generator.ts
@@ -79,10 +79,14 @@ export class Generator {
     const parameters = parameter.split(',');
     for (let param of parameters) {
       // remove double quote
-      param = param.substring(1, param.length - 1);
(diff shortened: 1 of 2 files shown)


src/generator.ts
```diff
@@ -79,10 +79,14 @@ export class Generator {
     const parameters = parameter.split(',');
     for (let param of parameters) {
       // remove double quote
-      param = param.substring(1, param.length - 1);
+      if (param[0] === '"' && param[param.length - 1] === '"') {
```
vam-google, line 82: is parm.lenght guaranteed to be >=1?

alexander-fenster: It does not matter, it won't crash in JavaScript (will return `undefined`).

PR #587, opened 2020-07-08
refactor: do not rebuild protos

alexander-fenster:
Background: the generator itself needs to use some protos files during its runtime. Previously, we compiled those proto files (using `pbjs` and `pbts` from protobuf.js) during the `bazel build` step, which proved to be flaky because of a LOT of extra complexity in protobuf.js bazel support (let's start from saying that this support comes from experimental `@bazel/labs` NPM package).

To reduce complexity of the client libraries, we tend to commit the compiled JS/TS files for protos into the repository (effectively making them generated offline when a developer runs `pbjs` / `pbts` manually). Let's do the same here.

Pros: 
* faster and more stable build;
* less dependencies (no need to depend on the experimental `@bazel/labs`).

Cons: 
* need to run `npm run compile-protos` if the protos (`google/api/*.proto`, or `google/longrunning/operations.proto`, or `google/protobuf/**/*.proto`) are changed.
  * This happens rarely. I added an NPM script for that, similar to the one we have in each library.


JustinBeckwith (approved):
I barely understand what's happening, but I trust you 😆

PR #596, opened 2020-07-10
fix: do not log parameters

alexander-fenster:
Since we completed the Bazel migration, we don't need to log these parameters anymore. This logging makes the Kokoro build logs too verbose.


vam-google (approved):
LGTM

PR #641, opened 2020-08-11
fix: properly use method descriptors in the generated client

alexander-fenster:
Apparently, the batching implementation in gapic-generator-typescript never actually worked because of two things:
- wrong (not camelized) method name in the descriptors object;
- the batching descriptor was never passed to gax.

This was reported by a customer in https://github.com/googleapis/nodejs-logging/issues/835. 

Looks like we don't have any test coverage for batching functionality in logging (I submitted https://github.com/googleapis/nodejs-logging/issues/870). This kind of explains the problem :) if a feature is not covered by any working integration test, we can never be sure the feature actually works.

This PR is the first step (making it work properly in the generator). The next step would be to cover this by a test in logging library.

### DIFF
--- a/baselines/asset/src/v1/asset_service_client.ts.baseline
+++ b/baselines/asset/src/v1/asset_service_client.ts.baseline
@@ -38,7 +38,12 @@ export class AssetServiceClient {
   private _protos: {};
   private _defaults: {[method: string]: gax.CallSettings};
   auth: gax.GoogleAuth;
-  descriptors: Descriptors = {page: {}, stream: {}, longrunning: {}, batching: {}};
+  descriptors: Descriptors = {
+    page: {},
+    stream: {},
+    longrunning: {},
+    batching: {},
+  };
   innerApiCalls: {[name: string]: Function};
   pathTemplates: {[name: string]: gax.PathTemplate};
   operationsClient: gax.OperationsClient;
@@ -232,12 +237,13 @@ export class AssetServiceClient {
           throw err;
         });
 
+      const descriptor =
+        this.descriptors.longrunning[methodName] ||
+        undefined;
       const apiCall = this._gaxModule.createApiCall(
         callPromise,
         this._defaults[methodName],
-        this.descriptors.page[methodName] ||
-            this.descriptors.stream[methodName] ||
-            this.descriptors.longrunning[methodName]
+        descriptor
       );
 
       this.innerApiCalls[methodName] = apiCall;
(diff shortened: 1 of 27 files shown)


baselines/logging/src/v2/logging_service_v2_client.ts.baseline
```diff
@@ -236,7 +241,7 @@ export class LoggingServiceV2Client {
     // requests; denote this.
 
     this.descriptors.batching = {
-      WriteLogEntries: new this._gaxModule.BundleDescriptor(
+      writeLogEntries: new this._gaxModule.BundleDescriptor(
```
alexander-fenster, line 244: FYI: this fixes the first part of the bug

baselines/logging/src/v2/logging_service_v2_client.ts.baseline
```diff
@@ -301,12 +306,14 @@ export class LoggingServiceV2Client {
           throw err;
         });
 
+      const descriptor =
+        this.descriptors.page[methodName] ||
+        this.descriptors.batching[methodName] ||
```
alexander-fenster, line 311: FYI: this fixes the second part of the bug (note that the batching descriptor was never passed down to `createApiCall` before).

alexander-fenster:
I added comments to point to the changes that actually fix the logging problem.

xiaozhenliu-gg5 (commented):
Thanks for fixing this! That was part of my batching task. :)

xiaozhenliu-gg5 (approved):

alexander-fenster:
@xiaozhenliu-gg5 No blaming here - after all, I reviewed and approved that code :) It's just one more example of why my approach - if it's not covered by tests, it does not work - is right. I encourage you to apply this approach for every project you're working on :)

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -232,12 +237,13 @@ export class AssetServiceClient {
...
+      const descriptor =
+        this.descriptors.longrunning[methodName] ||
+        undefined;
       const apiCall = this._gaxModule.createApiCall(
         callPromise,
         this._defaults[methodName],
-        this.descriptors.page[methodName] ||
-            this.descriptors.stream[methodName] ||
-            this.descriptors.longrunning[methodName]
+        descriptor
```
bcoe, line 246: was the bug simply that this `descriptor` value wasn't be passed into `createApiCall`?

bcoe (approved):
I think we'll want to be careful rolling this out, as a bunch of client libraries will suddenly behaving differently.

This looks like a great fix to me though.

baselines/asset/src/v1/asset_service_client.ts.baseline
alexander-fenster: the `this.descriptors.batching[methodName]` was not passed.

PR #667, opened 2020-08-20
fix: allow trailing segments after version in package name

alexander-fenster:
Ads protos have all services residing under `"google.ads.googleads.v4.services"` (see [here](https://github.com/googleapis/googleapis/blob/master/google/ads/googleads/v4/services/account_budget_proposal_service.proto#L17)) and having this `services` segment after the version apparently breaks the current naming heuristics.

Fixing that so that whenever a segment matching a possible version is found, everything after it is ignored for the naming purposes.

Cc: @lukesneeringer (since I largely taken the naming logic from your original Python implementation)
Cc: @aohren - FYI, this is the first of 2 PRs that will make Ads TypeScript work. The second one is coming.

### DIFF
--- a/src/schema/naming.ts
+++ b/src/schema/naming.ts
@@ -56,20 +56,37 @@ export class Naming {
     if (!invalidPrefix) {
       rootPackage = prefix.replace(/\.$/, '');
     }
+
     const segments = rootPackage.split('.');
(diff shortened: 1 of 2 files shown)


xiaozhenliu-gg5 (approved):
Thanks for the fix!

PR #668, opened 2020-08-20
test: add a baseline test that covers possible naming conflicts

alexander-fenster:
This PR adds a baseline test for one Very Bad Proto. This proto is so bad that the generated library will never compile (I left some comments below to show the problems).

We have some naming problems in the generator that do not affect any existing API so far, but one of those problem (the `*Stream` one, see below) affects Ads APIs and will backfire when we start generating Ads TypeScript client library.

Currently, the client library class has some extra stuff that can have name conflicts with protobuf methods:

1. For any paginated method `method`, we generate extra functions `methodStream` and `methodAsync`, both can conflict with an actual RPC with the same name.  (this is what happens with Ads protos - they have a paginated `rpc Something` and also `rpc SomethingStream`)

2. For any long running method `method`, we generate an extra function `checkMethodProgress` which can conflict with an actual RPC with the same name.

3. We have some other methods in our class (e.g. `initialize` and `getProjectId`) which will conflict with RPCs with the same names.

4. An RPC can be named in a way that won't work in TypeScript (e.g. `rpc Constructor` or `rpc Function`). This is discouraged by AIPs but possible.

As a result, the generator works perfectly in these cases, but the resulting code cannot compile.

This proto does not currently handle all bad use cases (e.g. the one in googleapis/google-cloud-node-core#645 is not covered) but I will update it with more naming conflict findings.

These problems are going to be solved by some naming refactor (I'm still thinking about the proper design), and when the code for the `naming` baseline finally compiles, I'll add its compilation to CircleCI to make we don't break things after this is fixed.

Cc: @noahdietz - as we discussed in the chat, it might be only TypeScript problem for now, but I guess we still can have problems with reserved words in other languages. Do you think making this kind of "bad bad proto" (covering all languages and not just TypeScript) would make sense as a team effort?


baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+ *
+ * @param {Object} request
+ *   The request object that will be sent.
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [Empty]{@link google.protobuf.Empty}.
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+ */
+  initialize(
```
alexander-fenster, line 552: oops! we already have `initialize` above

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+      options = {};
+    }
+    else {
+      options = optionsOrCallback as gax.CallOptions;
+    }
+    options = options || {};
+    this.initialize();
+    return this.innerApiCalls.initialize(request, options, callback);
+  }
+  servicePath(
```
alexander-fenster, line 579: oops! we already have `servicePath` above

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+ *
+ * @param {Object} request
+ *   The request object that will be sent.
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [Empty]{@link google.protobuf.Empty}.
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+ */
+  servicePath(
```
alexander-fenster, line 609: oops! we already have `servicePath` above

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+ *
+ * @param {Object} request
+ *   The request object that will be sent.
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [Empty]{@link google.protobuf.Empty}.
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+ */
+  apiEndpoint(
```
alexander-fenster, line 666: oops! we already have `apiEndpoint` above, line 250

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+ *
+ * @param {Object} request
+ *   The request object that will be sent.
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [Empty]{@link google.protobuf.Empty}.
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+ */
+  port(
```
alexander-fenster, line 723: oops! we already have `servicePath` above, line 257

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+ *
+ * @param {Object} request
+ *   The request object that will be sent.
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [Empty]{@link google.protobuf.Empty}.
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+ */
+  scopes(
```
alexander-fenster, line 780: oops! we already have `scopes` above, line 265

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+ *
+ * @param {Object} request
+ *   The request object that will be sent.
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [Empty]{@link google.protobuf.Empty}.
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+ */
+  getProjectId(
```
alexander-fenster, line 837: oops! we already have `getProjectId` above, line 276

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+ *
+ * @param {Object} request
+ *   The request object that will be sent.
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [Empty]{@link google.protobuf.Empty}.
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+ */
+  function(
```
alexander-fenster, line 895: Just oops

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+ *
+ * @param {Object} request
+ *   The request object that will be sent.
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Promise} - The promise which resolves to an array.
+ *   The first element of the array is an object representing [Empty]{@link google.protobuf.Empty}.
+ *   The promise has a method named "cancel" which cancels the ongoing API call.
+ */
+  constructor(
```
alexander-fenster, line 952: Big ooops

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1074 @@
...
+ *   The decoded operation object has result and metadata field to get information from.
+ *
+ * @example:
+ *   const decodedOperation = await checkLongRunningProgress(name);
+ *   console.log(decodedOperation.result);
+ *   console.log(decodedOperation.done);
+ *   console.log(decodedOperation.metadata);
+ *
+ */
+  async checkLongRunningProgress(name: string): Promise<LROperation<protos.google.protobuf.Empty, protos.google.protobuf.Empty>>{
```
alexander-fenster, line 1052: Oops! Line 494 already has this one.

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1168 @@
...
+ *   The request object that will be sent.
+ * @param {google.protobuf.Empty} request.content
+ * @param {number} request.pageSize
+ * @param {string} request.pageToken
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Stream}
+ *   An object stream which emits an object representing [Empty]{@link google.protobuf.Empty} on 'data' event.
+ */
+  paginatedMethodStream(
```
alexander-fenster, line 1107: Name conflict again.

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -0,0 +1,1168 @@
...
+ *   The request object that will be sent.
+ * @param {google.protobuf.Empty} request.content
+ * @param {number} request.pageSize
+ * @param {string} request.pageToken
+ * @param {object} [options]
+ *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
+ * @returns {Object}
+ *   An iterable Object that conforms to @link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols.
+ */
+  paginatedMethodAsync(
```
alexander-fenster, line 1137: Name conflict as well.

xiaozhenliu-gg5 (approved):
LGTM with some thoughts: 
1. `checkMethodProgress` for LRO methods, is it doing the same thing as our added `checkMethodProgress` helper? If it is, we probably wanna replace it, if not, we can change to another name for avoiding conflict.
2. when the rpc name hits the reserved words like `function`, should we throw error for them during generation? Otherwise, if we put this naming.proto to gapic-showcase, we probably wanna add more reserved words for other languanges like `class, enum, etc`. @noahdietz 
Thank you!

alexander-fenster:
@xiaozhenliu-gg5 re: `checkMethodProgress` - just imagine that this is some completely unrelated RPC in the proto, that happens to be in conflict with what we generate.  Similar to how `*Stream` or `*Async` will be in conflict if a method with this name exists - this is what actually happened :) 

@xiaozhenliu-gg5 @noahdietz re: "bad names", unfortunately, the corresponding AIP is not strict enough - [AIP-140](https://google.aip.dev/140) only suggests that _field_ names *should not* use words that might be reserved, and there is no guidance on method names. So all examples above are perfectly valid from AIP point of view.

What I'm planning to do with this is to introduce a simple method renaming logic in the generator - basically, the generator should know of the reserved names (all RPC names and language keywords) and it should not add conflicting names to the class. I expect to make the generator finally generate a working library based on this bad proto (which I consider to be AIP-compliant), but this is not the main goal here.

The main goal is to have a test coverage for this planned refactor. Before I can start working on this refactor, I must have a baseline test checked in, because it's the only way to cover the refactored code with tests (since _all_ our existing tests do not have naming conflicts, the name resolution logic will never trigger in the regular test coverage). Hence this new baseline.

I started this because of an Ads method that triggered the naming conflict similar to #1 above, and with this baseline, the future change to `*Stream` method will be visible. I may or may not fix other problems listed, but at least we'll have a test that shows the current behavior.

PR #682, opened 2020-08-28
feat: reduce possibility of name conflicts

alexander-fenster:
This PR introduces a _namer_ that, being a part of the template, makes sure that the generated extra method (such as `FooAsync` for method `Foo`, or `checkFooProgress` for long running method `Foo`) do not conflict with existing RPCs having the same names.

Implementation: a template may provide a JS file `namer.js` that will be loaded by the generator during its runtime and passed back to the template, so it could use the logic in JS to resolve name conflicts.

Let's assume the proto has a long running RPC `Foo`. Previously, we would've generated an extra method `checkFooProgress`, and if the proto has an actual RPC `CheckFooProgress`, the two names would've got into the conflict. With this change, the namer will detect the conflict, and the generated extra method will be called `checkFooProgress1` (2, 3, etc.) The same will happen if the proto has methods `Initialize`, `GetProjectId`, etc. (which can conflict with the methods added by the generator).

To use the namer in the template, use `{{ id.get("name") }}` instead of just `{{ name }}`.

Note that only "naming" baseline (a special proto with a lot of naming conflicts) is changed. No existing API should be affected by this feature.

This feature will unblock generation of Ads TypeScript library. Cc: @aohren

Small note: I removed the part of the `naming.proto` that checks for TypeScript keywords conflicts, I'll get to that part later in a separate PR.

### DIFF
--- a/baselines/naming/protos/google/naming/v1beta1/naming.proto.baseline
+++ b/baselines/naming/protos/google/naming/v1beta1/naming.proto.baseline
@@ -112,21 +112,6 @@ service Naming {
       body: "*"
     };
   }
-
-  // Problem #4: RPCs that conflict with TypeScript reserved words
-  rpc Function(google.protobuf.Empty) returns (google.protobuf.Empty) {
-    option (google.api.http) = {
-      post: "/v1beta1/naming:function"
-      body: "*"
-    };
-  }
-
-  rpc Constructor(google.protobuf.Empty) returns (google.protobuf.Empty) {
-    option (google.api.http) = {
-      post: "/v1beta1/naming:constructor"
-      body: "*"
-    };
-  }
 }
 
 message PaginatedMethodRequest {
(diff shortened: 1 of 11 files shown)


baselines/naming/protos/google/naming/v1beta1/naming.proto.baseline
```diff
@@ -112,21 +112,6 @@ service Naming {
       body: "*"
     };
   }
-
```
alexander-fenster, line 115: FYI: the lines removed from the test proto correspond to a feature (reserved words) that I'm not going to cover soon.

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -807,124 +807,9 @@ export class NamingClient {
       options = optionsOrCallback as gax.CallOptions;
     }
     options = options || {};
-    this.initialize();
+    this.initialize1();
     return this.innerApiCalls.getProjectId(request, options, callback);
   }
-  function(
```
alexander-fenster, line 813: FYI: The lines removed from the test baseline correspond to a feature (reserved words) that I'm not going to cover soon.

xiaozhenliu-gg5 (approved):

PR #696, opened 2020-09-03
fix: do not overwrite options object

### LINKED ISSUE #381: Fails to resolve GRPC endoint when called from GKE pod
#### Environment details

  - OS: Debian inside K8s pod
  - Node.js version: 10.15
  - npm version: 6.4.1
  - `@google-cloud/kms` version:2.1.2

#### Steps to reproduce

  1. Deployed pod with mounted certificate
  2. Make a client with said certificate
  3. Attempt to use any function of the client.

The following error will be produced: `"The GRPC target is not implemented on the server, host: secretmanager.googleapis.com, method: /google.cloud.kms.v1.KeyManagementService/Decrypt." ` with the final GRPC method being the issue.

NOTE: The cert was tested locally and works correctly, and works for its other IAM scopes. Also, the code in question is being called from a handler for a gRPC server, if that's any help. OpenTelemetry is also running in the process as well.

alexander-fenster:
Fixes https://github.com/googleapis/nodejs-kms/issues/381, https://github.com/googleapis/google-cloud-node-core/issues/636.

This change to the generator will be applied to all the generated libraries. 

The main change improves the handling of `opts` object in the generated client constructor (making sure that it never changes the object passed as a parameter). Using `??` and `?.` operators improved the code a little bit.

Another important change is to use TypeScript [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to make sure that the main client is exported both as a class and as a type (to allow both `const client = new FooClient()` and `let client: FooClient`, the latter does not work without this change.

Also, made some smaller changes here and there to improve the generated code (please see the baseline changes).

All changes are supposed to be compatible and non-breaking.

### DIFF
--- a/baselines/asset/src/v1/asset_service_client.ts.baseline
+++ b/baselines/asset/src/v1/asset_service_client.ts.baseline
@@ -74,12 +74,7 @@ export class AssetServiceClient {
    * @param {string} [options.apiEndpoint] - The domain name of the
    *     API remote host.
    * @param {gax.ClientConfig} [options.clientConfig] - client configuration override.
-   *     Client configuration set in the JSON config
-   *     (`src/v1/asset_service_client_config.json`)
-   *     can be overridden by passing `clientConfig` to the client constructor.
-   *     The detailed structure of the clientConfig can be found
-   *     {@link https://github.com/googleapis/gax-nodejs/blob/master/src/gax.ts#L564-L611 here}.
-   *     Pass any subset of that JSON config as `options.clientConfig`.
+   *     TODO(@alexander-fenster): link to gax documentation.
    * @param {boolean} fallback - Use HTTP fallback mode.
    *     In fallback mode, a special browser-compatible transport implementation is used
    *     instead of gRPC transport. In browser context (if the `window` object is defined)
@@ -90,9 +85,9 @@ export class AssetServiceClient {
   constructor(opts?: ClientOptions) {
     // Ensure that options include all the required fields.
     const staticMembers = this.constructor as typeof AssetServiceClient;
-    const servicePath = opts?.servicePath ?? opts?.apiEndpoint ?? staticMembers.servicePath;
-    const port = opts?.port ?? staticMembers.port;
-    const scopes = opts?.scopes ?? staticMembers.scopes;
+    const servicePath = opts?.servicePath || opts?.apiEndpoint || staticMembers.servicePath;
+    const port = opts?.port || staticMembers.port;
+    const scopes = opts?.scopes || staticMembers.scopes;
     const clientConfig = opts?.clientConfig ?? {};
     const fallback = opts?.fallback ?? (typeof window !== "undefined");
     opts = Object.assign({servicePath, port, scopes, clientConfig, fallback}, opts);
--- a/baselines/asset/src/v1/asset_service_client.ts.baseline
+++ b/baselines/asset/src/v1/asset_service_client.ts.baseline
@@ -73,40 +73,34 @@ export class AssetServiceClient {
    *     your project ID will be detected automatically.
    * @param {string} [options.apiEndpoint] - The domain name of the
    *     API remote host.
+   * @param {gax.ClientConfig} [options.clientConfig] - client configuration override.
+   *     Client configuration set in the JSON config
+   *     (`src/v1/asset_service_client_config.json`)
+   *     can be overridden by passing `clientConfig` to the client constructor.
+   *     The detailed structure of the clientConfig can be found
+   *     {@link https://github.com/googleapis/gax-nodejs/blob/master/src/gax.ts#L564-L611 here}.
+   *     Pass any subset of that JSON config as `options.clientConfig`.
+   * @param {boolean} fallback - Use HTTP fallback mode.
+   *     In fallback mode, a special browser-compatible transport implementation is used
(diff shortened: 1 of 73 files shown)


baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -73,40 +73,34 @@ export class AssetServiceClient {
    *     your project ID will be detected automatically.
    * @param {string} [options.apiEndpoint] - The domain name of the
    *     API remote host.
+   * @param {gax.ClientConfig} [options.clientConfig] - client configuration override.
+   *     Client configuration set in the JSON config
+   *     (`src/v1/asset_service_client_config.json`)
+   *     can be overridden by passing `clientConfig` to the client constructor.
+   *     The detailed structure of the clientConfig can be found
+   *     {@link https://github.com/googleapis/gax-nodejs/blob/master/src/gax.ts#L564-L611 here}.
```
JustinBeckwith, line 81: I see a few problems here.  I don't think the reference here will stay up to date as the code in `gax.ts` changes and moves.  So pinning to `master` is unlikely to stay correct.  Second - should this be a reference to another interface?  I'm trying to look a lot closer at the reference API documentation we generate.

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -73,40 +73,34 @@ export class AssetServiceClient {
...
-    //
-    // const customConfig = {"interfaces": {"google.showcase.v1beta1.Echo": {"methods": {"Echo": {"retry_codes_name": "idempotent", "retry_params_name": "default"}}}}}
-    // const showcaseClient = new showcaseClient({ projectId, customConfig });
-    opts.clientConfig = opts.clientConfig || {};
-
-    // If we're running in browser, it's OK to omit `fallback` since
-    // google-gax has `browser` field in its `package.json`.
-    // For Electron (which does not respect `browser` field),
-    // pass `{fallback: true}` to the AssetServiceClient constructor.
+    const servicePath = opts?.servicePath ?? opts?.apiEndpoint ?? staticMembers.servicePath;
```
JustinBeckwith, line 93: I'm being wildly pedantic, but this could be considered a breaking change.  `??` has subtle different behavior from `||` in the way they treat falsey non-null/undefined values.  In practice, this means if the `opts.servicePath` was set to `''`, this would be a behavioral change.  I defer to @bcoe to tell us how much this matters.

baselines/asset/src/v1/asset_service_client.ts.baseline
alexander-fenster: Yeah it can move. I need to see if I can make a proper jsdoc link to a gax interface.

baselines/asset/src/v1/asset_service_client.ts.baseline
alexander-fenster: It should be safe to use `||` here, right?  if I write `opts?.servicePath || opts?.apiEndpoint || staticMembers.servicePath`, it will be the same as before, but still much more readable. WDYT?

JustinBeckwith: I tend to agree.

bcoe: I thinking the optional chain with the || is a fine compromise. I feel like someone might expect an empty string to default to the API endpoint.

baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
```diff
@@ -74,40 +74,34 @@ export class BigQueryStorageClient {
    *     your project ID will be detected automatically.
    * @param {string} [options.apiEndpoint] - The domain name of the
    *     API remote host.
+   * @param {gax.ClientConfig} [options.clientConfig] - client configuration override.
+   *     Client configuration set in the JSON config
+   *     (`src/v1beta1/big_query_storage_client_config.json`)
+   *     can be overridden by passing `clientConfig` to the client constructor.
+   *     The detailed structure of the clientConfig can be found
```
bcoe, line 81: Why don't we just pull this detailed configuration into the README, if we think it's important to call out, that way the link won't change.

bcoe (changes requested):
A couple nits.

baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
alexander-fenster: I have removed all this docs stuff from this PR. As discussed, I'll prepare a separate markdown file in google-gax and will update all docs in a separate generator PR (today or next week).

baselines/asset/src/v1/asset_service_client.ts.baseline
alexander-fenster: Done!

baselines/asset/src/v1/asset_service_client.ts.baseline
alexander-fenster: I removed all links to gax in this PR and will take care of this separately (as discussed, will add links to a separate document).

bcoe (approved):

JustinBeckwith (approved):

PR #760, opened 2020-10-30
fix(docs): better jsdoc for generated libraries

alexander-fenster:
This PR improves the generated documentation and adds a very simple JSDoc examples as well as links to https://github.com/googleapis/gax-nodejs/blob/master/client-libraries.md.


baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -52,36 +52,10 @@ export class AssetServiceClient {
...
-   * @param {gax.ClientConfig} [options.clientConfig] - client configuration override.
-   *     TODO(@alexander-fenster): link to gax documentation.
-   * @param {boolean} fallback - Use HTTP fallback mode.
-   *     In fallback mode, a special browser-compatible transport implementation is used
-   *     instead of gRPC transport. In browser context (if the `window` object is defined)
-   *     the fallback mode is enabled automatically; set `options.fallback` to `false`
-   *     if you need to override this behavior.
+   * @param {object} [options] - The configuration object.
+   * The options accepted by the constructor are described
+   * in [this document](https://github.com/googleapis/gax-nodejs/blob/master/client-libraries.md#creating-the-client-instance).
```
JustinBeckwith, line 57: I don't think this is a good change.  This would remove all of the interesting content from this page:
https://googleapis.dev/nodejs/vision/latest/v1p1beta1.ImageAnnotatorClient.html

And replace it with a link to something in a README that is disconnected from the generated code.  In the next ~6 months, we are going to start using TypeScript types to infer some of this stuff from the interface definitions.  I'm curious - what was the motivation for this change?

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -351,7 +328,11 @@ export class AssetServiceClient {
  *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
  * @returns {Promise} - The promise which resolves to an array.
  *   The first element of the array is an object representing [BatchGetAssetsHistoryResponse]{@link google.cloud.asset.v1.BatchGetAssetsHistoryResponse}.
- *   The promise has a method named "cancel" which cancels the ongoing API call.
+ *   Please see the
+ *   [documentation](https://github.com/googleapis/gax-nodejs/blob/master/client-libraries.md#regular-methods)
```
JustinBeckwith, line 332: @fhinkel @tbpg I think we need to spend some time thinking through what we want generators to do here :)

JustinBeckwith (changes requested):

bcoe (commented):
Will defer to @fhinkel and @JustinBeckwith on this one, since we're in the midst of moving over our doc generation to a new approach.

I do think it would be good to figure out how to go about thoroughly documenting behavior that's common to all generated libraries:

* HTTPs fallbacks.
* async iteration.
* proxy support.

baselines/asset/src/v1/asset_service_client.ts.baseline
alexander-fenster: The main reason here is to be able to add options / fix documentation without regenerating 50 libraries. We can probably list just a few most important options here, and refer to the external document for more information.

alexander-fenster: I put it back!

baselines/asset/src/v1/asset_service_client.ts.baseline
alexander-fenster: Here I really want to stop advertising the useless `.cancel()` method in hope that we may be able to drop it later, like in 2 years or so :)

JustinBeckwith (approved):

PR #761, opened 2020-11-04
feat: set defaultScopes in auth client

alexander-fenster:
If a client is connecting to the default endpoint, set `defaultScopes` and do not set `scopes`.
Set `scopes` otherwise.

### DIFF
--- a/baselines/asset/src/v1/asset_service_client.ts.baseline
+++ b/baselines/asset/src/v1/asset_service_client.ts.baseline
@@ -88,10 +88,14 @@ export class AssetServiceClient {
     const staticMembers = this.constructor as typeof AssetServiceClient;
     const servicePath = opts?.servicePath || opts?.apiEndpoint || staticMembers.servicePath;
     const port = opts?.port || staticMembers.port;
-    const scopes = opts?.scopes || staticMembers.scopes;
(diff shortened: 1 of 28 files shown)


baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -105,6 +109,11 @@ export class AssetServiceClient {
     // Save the auth object to the client, for use by other methods.
     this.auth = (this._gaxGrpc.auth as gax.GoogleAuth);
 
+    // Set the default scopes in auth client if needed.
+    if (servicePath === staticMembers.servicePath) {
```
bcoe, line 113: that's a nice simple approach 👍 it seems like enough of an edge case that someone would purposefully override `sevicePath`, with the same value, that I'd say let's go with this approach and cross that bridge if it ever happens.

bcoe (approved):
this looks good to me 👍 we'll also need to make sure we've bumped the gax version as we take this change, and that gax enforces the latest version of auth?

alexander-fenster:
Yes, you're right about the latest gax. I'm working on fixing Node 15 tests there after that I will be able to make a release that is guaranteed to fetch the latest auth library.

PR #763, opened 2020-11-06
fix: use yarn instead of npm

alexander-fenster:
Unfortunately for us @bcoe, @vam-google and I figured that with NPM bazel rule each subsequent run (when all the node modules are already installed) takes 11 extra seconds compared with 1.5 seconds with yarn.  In `synthtool` context, this will add more than 10 minutes to the nighly process for no reason.  I suggest we rollback to `yarn_install` until this extra delay is fixed (either in npm or on bazel rules).


vam-google:
LGTM, but I defer to the TS experts for the "actual" LGTM.

bcoe (approved):

WORKSPACE
```diff
@@ -18,10 +18,10 @@ load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies", "rules_
 rules_proto_dependencies()
 rules_proto_toolchains()
 
-load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories", "npm_install")
```
bcoe, line 21: could this rule be using `npm ci`?

alexander-fenster: https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/npm_install/npm_install.bzl#L211 - they don't have it, but I guess I need to investigate if it can be done better there.

PR #803, opened 2020-12-15
feat: DIREGAPIC generation

alexander-fenster:


alexander-fenster:
Not needed since all fixes are already merged.

PR #826, opened 2021-03-04
refactor: make it work with latest Bazel rules

alexander-fenster:
1. Upgraded Bazel Node.js rules to v3.2.1 and made sure the generator actually works when called from `googleapis`
2. Removed `module-alias` hack (required moving stuff around but now there's no magic in the build process)
3. Made it possible to pass `protoc` location as an option for a generator wrapper (used only in unit tests; `googleapis` process calls the protoc plugin directly so is not affected).


xiaozhenliu-gg5 (approved):
Thanks for doing this!

PR #845, opened 2021-03-23
feat: api_dump template to use in tests

alexander-fenster:
Before fixing the naming issue for Pub/Sub, adding some test coverage.

The generator now has a template called `api_dump` which just dumps the JSON representation of the API metadata being generated. As you can see, the package name for the Pub/Sub API is currently generated as `google` and not `google.pubsub`, which is the source of the Pub/Sub problems, and it will be fixed in the following PR.

Please ignore the copyright year changes in the baseline protos.


baselines/pubsub-api-dump/api.json.baseline
```diff
@@ -0,0 +1,14065 @@
+{
+  "packageName": "google",
```
alexander-fenster, line 2: This is the problem I will fix in the following PR: the package name must be `google.pubsub` but is detected incorrectly.

templates/api_dump/api.json.njk
```diff
@@ -0,0 +1 @@
+{{ api.dump | safe }}
```
alexander-fenster, line 1: This is the new template being added for test purposes. Running the generator with `--template=api_dump` will produce the above JSON representation, which will help debugging weird problems.

typescript/src/schema/api.ts
```diff
@@ -193,6 +195,17 @@ export class API {
       '  '
     );
   }
+
+  get dump() {
```
alexander-fenster, line 199: This is a new getter to produce an API dump.

alexander-fenster: Just to clarify, it will be only used in tests or during debugging. We won't dump this thing for regular libraries.

bcoe (approved):

baselines/asset/protos/google/cloud/asset/v1/asset_service.proto.baseline
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 Google LLC
```
bcoe, line 1: This is a little strange that it changes the date?

alexander-fenster: The original protos are (c) 2020 and apparently someone changed _everything_ to 2021 even though this particular change was not needed.

PR #847, opened 2021-03-25
fix: name generation when multiple packages are involved

alexander-fenster:
Fixes the problem with incorrect package name in Pub/Sub library (and possibly others), now with the test coverage (hence the API JSON change). After this is released, it should generate a proper config JSON.

### DIFF
--- a/baselines/asset/src/v1/asset_service_proto_list.json.baseline
+++ b/baselines/asset/src/v1/asset_service_proto_list.json.baseline
@@ -1,9 +1,4 @@
 [
   "../../protos/google/cloud/asset/v1/asset_service.proto",
-  "../../protos/google/cloud/asset/v1/assets.proto",
-  "../../protos/google/cloud/orgpolicy/v1/orgpolicy.proto",
-  "../../protos/google/identity/accesscontextmanager/type/device_resources.proto",
-  "../../protos/google/identity/accesscontextmanager/v1/access_level.proto",
-  "../../protos/google/identity/accesscontextmanager/v1/access_policy.proto",
-  "../../protos/google/identity/accesscontextmanager/v1/service_perimeter.proto"
+  "../../protos/google/cloud/asset/v1/assets.proto"
 ]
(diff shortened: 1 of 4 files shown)


baselines/asset/src/v1/asset_service_proto_list.json.baseline
```diff
@@ -1,9 +1,4 @@
 [
   "../../protos/google/cloud/asset/v1/asset_service.proto",
-  "../../protos/google/cloud/asset/v1/assets.proto",
```
alexander-fenster, line 3: This won't change anything in the generated code.

typescript/src/generator.ts
```diff
@@ -181,15 +181,12 @@ export class Generator {
...
-    const protoFilesToGenerate = this.request.protoFile.filter(
-      pf =>
-        pf.name &&
-        this.request.fileToGenerate.includes(pf.name) &&
-        !API.isIgnoredService(pf)
-    );
-    const packageNamesToGenerate = protoFilesToGenerate.map(
-      pf => pf.package || ''
-    );
+    const protoPackagesToGenerate = new Set<string>();
```
alexander-fenster, line 184: This fix follows https://github.com/googleapis/gapic-generator-typescript/pull/832 - this is just one more place I forgot to fix (and did not catch it because of lack of test coverage).  Now this is covered by tests :)

typescript/src/generator.ts
```diff
@@ -181,15 +181,17 @@ export class Generator {
...
+    API.filterOutIgnoredServices(
+      this.request.protoFile.filter(
+        fd =>
+          this.request.fileToGenerate.includes(fd.name!) &&
+          fd.service &&
+          fd.service.length > 0
+      )
+    ).forEach(fd => protoPackagesToGenerate.add(fd.package || ''));
+    console.warn(protoPackagesToGenerate);
```
bcoe, line 193: Do you want to leave this debug message in?

bcoe (approved):

typescript/src/generator.ts
alexander-fenster: No! Thank you for spotting it.

PR #870, opened 2021-05-10
fix: the return type from IAM method promises should be an array

feywind:
This is a fix for this issue on nodejs-pubsub: https://github.com/googleapis/google-cloud-node-core/issues/387

@alexander-fenster This is the thing we talked about in chat :)


alexander-fenster (approved):

alexander-fenster:
I believe this needs to be reevaluated now that we support IAM mix-in. I'll close this for now.

feywind:
@alexander-fenster We're trying to get the gax side of this in for the new major - can we do that re-evaluation on this? I'm not sure what IAM mixins are.

Also @sofisl who's doing the Node 14 work.

feywind:
This PR is out of date now.

PR #876, opened 2021-05-19
Update service_client to append @deprecate tag

alicejli:
@alexander-fenster @summer-ji-eng 

Updated service_client to append @deprecate doc tag for services and methods.
For: https://github.com/googleapis/google-cloud-node-core/issues/622

TODO: Emit a DeprecationWarning


summer-ji-eng (changes requested):
Congratulations🥳 on your first PR.

One suggestion: could you add one simple baseline test for deprecation. It is also feasible for your next emit deprecation warning.

You might need to add two simple protos with marking the service deprecation option and method deprecate. ([example proto](https://github.com/googleapis/gapic-generator-java/pull/708/files#diff-20469219bfdf2257f72cc55e1d1dcbb1e93b73f927bc8f1c43a8565aaa7cd90a))
After setup baseline, you could run 'npm run docs' to see if the JSDoc contains the annotation.
Feel free to reach out if you have any question. Thank you.

alicejli:
Thanks for your feedback and patience! I've updated the template for rendering in jsdoc-fresh (https://github.com/googleapis/jsdoc-fresh/commit/def7b8280e52a943c519ac44b8534afa0b49a711), and updated this code to include a baseline test as well as to generate a warning. Let me know any other feedback, thank you!

summer-ji-eng (approved):
🚀 Thanks a lot adding the deprecate test and solving the displaying deprecate tag problem.

.github/workflows/ci.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
         echo "The old one will disappear after 7 days."
 
     - name: Run bazel build
-      run: bazel build '//...'
+      run: bazel test --test_output=errors //...
```
summer-ji-eng, line 45: Why do we change this line? Intentionally?

alicejli: Oops - that line was meant to replace line 48 (`bazel test`) so that it will print out the errors if the test fails. Originally, if a test fails, it's hard to debug since you cannot see the errors. Updated this file to keep line 45 the same, and updated line 48.

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -293,6 +296,9 @@ export class {{ service.name }}Client {
   {{ id.get("initialize") }}() {
     // If the client stub promise is already initialized, return immediately.
     if (this.{{ service.name.toCamelCase() }}Stub) {
+      {%- if service.options and service.options.deprecated %}
+      gax.warn('DEP1000', '{{service.name}} is deprecated and may be removed in a future version.', 'DeprecationWarning');
```
alexander-fenster, line 300: Can we have a service name in the ID? Let's say there are two deprecated services, we want to receive separate warnings from all of them (but once for each).  We can make the ID look like `DEP${service.name}` for example.

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -466,23 +478,35 @@ export class {{ service.name }}Client {
     }
     {{ util.buildHeaderRequestParam(method) }}
     this.{{ id.get("initialize") }}();
+    {%- if method.options and method.options.deprecated %}
+    gax.warn('DEP1001','{{method.name}} is deprecated and may be removed in a future version.', 'DeprecationWarning');
```
alexander-fenster, line 482: Same here, maybe `DEP${service.name}-${method.name}` or something?

summer-ji-eng: BTW, what's the DEP 1001 and above DEP 1000? Maybe I miss something here.

alicejli: I just made up the ID so there was a separate ID for a deprecated service and a deprecated method. But to Alex's point, they would not be unique for different services or methods, so using the service/method name in the ID itself makes more sense.

templates/typescript_gapic/src/$version/$service_client.ts.njk
alicejli: done

PR #877, opened 2021-05-19
fix: [diregapic]missing errors.proto

summer-ji-eng:
googleads protos are different than most other apis.
It has multiple packages, like 
```
package google.ads.googleads.v7.services;
package google.ads.googleads.v7.errors;
```
And errors.proto has no service.

Fix: we should allow package'name which has no service differ from service package name.

Local Test: generate googleads
```
$ bazel build //google/ads/googleads/v7:googleads-nodejs
$ cd .../bazel-bin/google/ads/googleads/v7
$ tar -xf googleads-nodejs.tar.gz
$ grep errors.proto googleads-nodejs/src/v7/*json | wc -l
```
>>>
Results: `126`


summer-ji-eng:
duplicate with #878

PR #886, opened 2021-05-31
feat: allow optional label for proto3

summer-ji-eng:
Enable `optional` presence tracking for proto3 messages. This flag is available in protobuf release >=3.12.
See more at [field_presence.md](https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md)


alexander-fenster:
We don't need it since protoc v3.15.0 :)

PR #888, opened 2021-06-02
fix: yargs update by passing interface to protect the type

summer-ji-eng:
With update on both package:
`yargs` -> `^16.2.0` -> `17.0.0`
`sinon` -> `^10.0.0` -> `^11.0.0`
see #865 

The typescript can't compile on type checks with error msg:
```
error TS2339: Property 'outputDir' does not exist on type '{ [x: string]: unknown; _: (string | number)[]; $0: string; } | Promise<{ [x: string]: unknown; _: (string | number)[]; $0: string; }>'.
  Property 'outputDir' does not exist on type 'Promise<{ [x: string]: unknown; _: (string | number)[]; $0: string; }>'.
  ```
because compiler expect `argv['outputDir'] rather than `argv.outputDir`.
The solution is pass in a interface to protect the type and avoid the typo by passing string in `[]`.

Once it merge, the #880 should pass the test.

### CODE REVIEW

File: typescript/src/gapic-generator-typescript.ts [hunk 1]
```diff
@@ -84,7 +84,27 @@ yargs.usage('Usage: $0 -I /path/to/googleapis');
...
+  packageName?: string | undefined;
+  mainService?: string | undefined;
+  template?: string | undefined;
+  gapicValidatorOut?: string | undefined;
+  validation?: string | undefined;
+  metadata?: boolean | undefined;
+  protoc?: string | undefined;
+  protoDirs?: string[];
+  commonProtoPath?: string;
+  descriptor: string | undefined;
```
alexander-fenster, line 102: This should be `?` too

typescript/src/gapic-generator-typescript.ts
```diff
@@ -84,7 +84,27 @@ yargs.usage('Usage: $0 -I /path/to/googleapis');
 yargs.usage('  --output_dir /path/to/output_directory');
 yargs.usage('  google/example/api/v1/api.proto');
 
-const argv = yargs.argv;
+export interface IArguments {
+  [x: string]: unknown;
+  outputDir?: string;
+  grpcServiceConfig?: string | undefined;
```
alexander-fenster, line 90: I think when you use `?` it already means "can be undefined" so should be enough to say just `grpcServiceConfig?: string` without `| undefined`.  Same for others.

alexander-fenster (approved):

typescript/src/gapic-generator-typescript.ts
summer-ji-eng: Thanks for catching this up.

typescript/src/gapic-generator-typescript.ts
summer-ji-eng: SG. remove `undefined`.

PR #891, opened 2021-06-05
feat(diregapic)(pagination): Add input option for google discovery api [pg1/3]

summer-ji-eng:
Add input option to distinguish the discovery api from googleapis. This is pre-PR for adding REGAPIC pagination feature. #887 

Usage Cases:
1. bazel rule pass "rest" parameters.
```
nodejs_gapic_library(
  name = "compute_small_nodejs_gapic",    
  package_name = "@google-cloud/compute-small",    
  src = ":compute_small_proto_with_info",    
  extra_protoc_parameters = ["metadata", "transport=rest"],    deps = [],
)
```
2. pass option in 
```
bazel run //:gapic_generator_typescript -- -I $GOOGLEAPIS_DISCOVERY --output-dir /tmp/output $GOOGLEAPIS_DISCOVERY/google/cloud/compute/v1/compute_small.proto --package-name @google-cloud/compute --transport rest
```

External User(No need pass `{fallback:'rest'}` in the clientOptions:
```
const compute = require('@google-cloud/compute');
const client = new compute.AddressesClient();
const project = await client.getProjectId();
```


templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -119,6 +119,14 @@ export class {{ service.name }}Client {
     const servicePath = opts?.servicePath || opts?.apiEndpoint || staticMembers.{{ id.get("servicePath") }};
     const port = opts?.port || staticMembers.{{ id.get("port") }};
     const clientConfig = opts?.clientConfig ?? {};
+    {% if api.rest -%}
+    // Implicitely set 'rest' value for the apis use rest as transport (eg. googleapis-discovery apis).
+    if (!opts) {
```
alexander-fenster, line 124: let's do `if (!(opts && 'fallback' in opts)) - so if the user actually passed `fallback` to the constructor we used whatever the user passed

alexander-fenster (approved):

PR #900, opened 2021-06-16
feat(mtls): pass variable to indicate whether default service endpoint used

bcoe:
In support of mtls work, pass variable to indicate whether or not the default service account has been used.

Refs: https://google.aip.dev/auth/4114


baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -31,6 +31,28 @@ import * as gapicConfig from './asset_service_client_config.json';
...
+/**
+ * Detect mTLS client certificate based on logic described in
+ * https://google.aip.dev/auth/4114.
+ *
+ * @param {object} [options] - The configuration object.
+ * @returns {Array} Array of strings representing client cert and key.
+ */
+function detectClientCertificate (opts?: ClientOptions) {
+  // If GOOGLE_API_USE_CLIENT_CERTIFICATE is true...:
+  if (process?.env?.GOOGLE_API_USE_CLIENT_CERTIFICATE === 'true') {
```
bcoe, line 43: Used `?.` to protect us in a browser environment, where `process.env` might be `undefined`.

bcoe:
CC: @arithmetic1728

baselines/asset/src/v1/asset_service_client.ts.baseline
alexander-fenster: Please add `typeof process !== 'undefined'`.  `process?.env?` will still fail if there is no process.

```
> pr0cess?.env?.node   // typo intended just to show how this would fail
Uncaught ReferenceError: pr0cess is not defined

> typeof pr0cess !== 'undefined' && pr0cess?.env?.node
false
```

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -92,14 +157,16 @@ export class AssetServiceClient {
   constructor(opts?: ClientOptions) {
     // Ensure that options include all the required fields.
     const staticMembers = this.constructor as typeof AssetServiceClient;
-    const servicePath = opts?.servicePath || opts?.apiEndpoint || staticMembers.servicePath;
+    // See: https://google.aip.dev/auth/4114
+    const [cert, key] = detectClientCertificate(opts);
```
alexander-fenster, line 161: I'm a little bit concerned about using synchronous file API in the constructor. We have an asynchronous `initialize()` method, maybe this logic can be moved there (using promisified version of `fs`)?

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -92,14 +157,16 @@ export class AssetServiceClient {
...
+    const servicePath = opts?.servicePath || opts?.apiEndpoint || this.mtlsServicePath(cert, key);
```
arithmetic1728, line 162: It looks like the logic should be:
```
const servicePath = opts?.apiEndpoint || this.mtlsServicePath(cert, key);
```
User provided endpoint always comes first.

arithmetic1728:
LGTM for the mTLS logic.

baselines/asset/src/v1/asset_service_client.ts.baseline
bcoe: This is fixed in most recently pushed code:

` `if (servicePath !== staticMembers.{{ id.get("servicePath") }}) return servicePath;`

baselines/asset/src/v1/asset_service_client.ts.baseline
bcoe: done.

baselines/asset/src/v1/asset_service_client.ts.baseline
bcoe: I've refactored to use async methods.

bcoe:
smoke test with existing library [here](https://github.com/googleapis/nodejs-pubsub/pull/1322).

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -31,6 +35,85 @@ import * as gapicConfig from './asset_service_client_config.json';
...
+    }
+    // If context aware metadata exists, run the cert provider command,
+    // parse the output to extract cert and key, and use this cert/key.
+    const metadataPath = join(
+      homedir(),
+      '.secureConnect',
+      'context_aware_metadata.json'
+    );
+    try {
+      const metadata = JSON.parse(await readFileAsync(metadataPath));
```
arithmetic1728, line 63: We only need to read the file if the metadata path exists. It is OK that the file doesn't exist, it is not an error or warning.

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -31,6 +35,85 @@ import * as gapicConfig from './asset_service_client_config.json';
...
+      );
+      const matchCert = stdout.toString().match(certRegex);
+      const matchKey = stdout.toString().match(keyRegex);
+      if (!(matchCert?.groups && matchKey?.groups)) {
+        throw Error('unable to parse certificate and key');
+      } else {
+        return [matchCert.groups.cert, matchKey.groups.key];
+      }
+    } catch (err) {
+      console.warn(
```
arithmetic1728, line 79: If there are any problems getting the cert/key, then it a real error, I think we don't need the try/catch block.

arithmetic1728 (approved):

arithmetic1728:
I think most of the mtls logic should be put in gax-nodejs instead of here. Then in generator, we just need to call the api provided by gax-nodejs.

summer-ji-eng:
> I think most of the mtls logic should be put in gax-nodejs instead of here. Then in generator, we just need to call the api provided by gax-nodejs.

@bcoe Thank you doing this. LGTM with mTLS logic. It might two things:
1. could you move the mtls logic into gax-nodejs to reduce client library surface code.
2. could you also add unit tests 🙏

baselines/asset/src/v1/asset_service_client.ts.baseline
bcoe: `readFileAsync` will throw on `ENOTFOUND` and we'll receive the following warning on the terminal:

```
failure loading context_aware_metadata.json: ENOENT: no such file or directory, open '/Users/bencoe/.secureConnectblerg/context_aware_metadata.json'
```

An attempt will be made to recover from this error and connect to the remote.

bcoe:
We should land https://github.com/googleapis/gax-nodejs/pull/1034 before merging.

> could you also add unit tests

Is it okay if I add the unit tests to `gax-nodejs` and we stick with the baseline tests in this repository, now that the logic has been reduced to setting a single private variable... I didn't see any existing unit tests in this codebase that I could build on.

bcoe:
@summer-ji-eng regarding unit tests, it looks like integration tests catch the additional parameter being added 👍

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -39,6 +39,7 @@ const version = require('../../../package.json').version;
 export class AssetServiceClient {
   private _terminated = false;
   private _opts: ClientOptions;
+  private _customServicePath: boolean;
```
alexander-fenster, line 42: nit: `_customServicePath` is probably not a good name for a _boolean_.  Maybe `_hasCustomServicePath` or `_needsCustomServicePath` or anything similar, starting with a verb?  Otherwise, `_customServicePath` reads as it's a string path.

alexander-fenster (approved):
LGTM with a small naming/style nit.

PR #901, opened 2021-06-16
fix: updated typescript version for compiler

alicejli:
When running the generator locally, bazel demands the typescript version to be `">=3.0.0 <4.3` which conflicts with the typescript version of `"^4.3.2` the compiler is currently using. 

Therefore, this change is to downgrade the typescript version used for compiling the generator locally in order for bazel rules to work. The generator does not actually use any 4.3+ features, so this downgrade should not affect any functionality (and does not affect the typescript version used for the generated client libraries, which is the latest one). 

We can bump this version back up once bazel is able to accept the latest version of typescript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "gts": "^3.1.0",
     "mocha": "^8.4.0",
     "sinon": "^11.1.1",
-    "typescript": "^4.3.2"
+    "typescript": "~4.2.0"
   },
   "engines": {
     "node": ">=v10.24.0"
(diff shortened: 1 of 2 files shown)


summer-ji-eng:
@alicejli can I have more context why do we need to downgrade `typescript` package, please?

alicejli:
> @alicejli can I have more context why do we need to downgrade `typescript` package, please?

Updated the original PR comment - let me know if that makes sense, thanks!

alexander-fenster (approved):

PR #902, opened 2021-06-16
fix(bazel): Remove monolith Bazel rule deps [typescript]

miraleung:
This is needed for the preliminary removal of the monolith rules from googleapis' `switched_rules_by_language`.

### DIFF
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -20,6 +20,14 @@ def gapic_generator_typescript_repositories():
       ],
   )
 
+  _rules_gapic_version = "0.5.4"
+  maybe(
+      http_archive,
+      name = "rules_gapic",
+      strip_prefix = "rules_gapic-%s" % _rules_gapic_version,
+      urls = ["https://github.com/googleapis/rules_gapic/archive/v%s.tar.gz" % _rules_gapic_version],
+  )
+ 
   maybe(
       http_archive,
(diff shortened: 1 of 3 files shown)


summer-ji-eng:
Thanks @miraleung help on this. LGTM, but please let @alexander-fenster take a quick look.
Can I have some context on `@com_google_api_codegen`. I know @rules_gapic lives at https://github.com/googleapis/rules_gapic.

Another question fro @alexander-fenster is that: 
We define our nodejs bazel rule at generator repo, [rules_typescript_gapic](https://github.com/googleapis/gapic-generator-typescript/tree/master/rules_typescript_gapic). But other lanugage Java, php, csharp all lives in [rules_gapic](https://github.com/googleapis/rules_gapic).
Try to understand the difference here. Thanks a lot.

summer-ji-eng (approved):

alexander-fenster:
The change LGTM.

@summer-ji-eng I think it makes more sense for the generator rules to be in the generator repo, because the rules and the generator are linked together (e.g. you can add a new option to the generator and to the rule implementation in the same PR). Other languages might do it in a different way though.

miraleung:
@summer-ji-eng +1 to @alexander-fenster. For the most part, other languages' rules also live in the generators, and those that don't should be cleaned up at some point.

PR #908, opened 2021-06-26
feat: add variables for self-signed auth to generator

sofisl:
In conjunction with: https://github.com/googleapis/google-auth-library-nodejs/pull/1196

### DIFF
--- a/baselines/asset/src/v1/asset_service_client.ts.baseline
+++ b/baselines/asset/src/v1/asset_service_client.ts.baseline
@@ -117,9 +117,8 @@ export class AssetServiceClient {
 
     // Save the auth object to the client, for use by other methods.
     this.auth = (this._gaxGrpc.auth as gax.GoogleAuth);
-
-    // Set useJWTAccessAlways on the auth object.
-    this.auth.useJWTAccessAlways = true;
+    // Set useJWTAccessWithScope on the auth object.
+    this.auth.useJWTAccessWithScope = true;
 
     // Set defaultServicePath on the auth object.
     this.auth.defaultServicePath = staticMembers.servicePath;
--- a/baselines/asset/src/v1/asset_service_client.ts.baseline
+++ b/baselines/asset/src/v1/asset_service_client.ts.baseline
@@ -118,6 +118,12 @@ export class AssetServiceClient {
     // Save the auth object to the client, for use by other methods.
     this.auth = (this._gaxGrpc.auth as gax.GoogleAuth);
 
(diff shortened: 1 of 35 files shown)


bcoe:
@sofisl, @busunkim96 surfaced an issue that means we shouldn't move ahead with this code change yet.

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -115,6 +115,12 @@ export class AssetServiceClient {
     // Save the auth object to the client, for use by other methods.
     this.auth = (this._gaxGrpc.auth as gax.GoogleAuth);
 
+    // Set useJWTAccessAlways on the auth object.
+    this.auth.useJWTAccessAlways = true;
```
alexander-fenster, line 119: Just to confirm: there is no opt out, it will be always set to `true`, the user cannot override this in any way. Is it how it's supposed to be?

alexander-fenster (approved):

baselines/asset/src/v1/asset_service_client.ts.baseline
bcoe: I can't think of a reason why a user would want to opt out. This variable is here to allow us to differentiate a GAPIC library from another library, e.g., sheets, as only the GAPIC libraries work with self signed JWT.

We should be mindful of testing this with DIREGAPIC libraries, now that I think of it 🤔

summer-ji-eng (approved):
LGTM, once you resolved the test. Thanks for working on this.

baselines/asset/src/v1/asset_service_client.ts.baseline
bshaffer: As long as this variable can be set to false for DIREGAPIC, then this should be sufficient

bshaffer (changes requested):
If it isn't too late, I'd like to advocate for all of these to be renamed from `useJWTAccessAlways` to `useJwtAccessWithScope` for consistency (see [AIP #4111](https://google.aip.dev/auth/4111#scope-vs-audience)), unless this was a conscious decision with justification.

bshaffer (approved):
As https://github.com/googleapis/google-auth-library-nodejs/pull/1204 is already rolled out and uses the name `useJWTAccessAlways`, we can keep that variable name

baselines/asset/src/v1/asset_service_client.ts.baseline
sofisl: We've added the DIREGAPIC variable now.

PR #909, opened 2021-06-28
feat!: Update IAMPolicy mixin to use <$service>.yaml file instead of explicit generator flag

alicejli:
As part of adding support generally for client library mixin method generations (draft AIP is at aip-dev/google.aip.dev#694), this update is to move the generator from reading the `google.iam.v1.IAMPolicy` mixin as an explicit generator flag to reading it from the <$service>.yaml file. 

This change should not impact functionality in any way, as each API that had the explicit mixin flag should also have it listed in their <$service>.yaml file. However, this *is* considered a breaking change to the generator so when this change is merged, it will bump the version of the generator.

Ping me for the internal doc(s) if desired.


baselines/pubsub-api-dump/api.json.baseline
```diff
@@ -495,7 +495,6 @@
         ".google.api.oauthScopes": "https://www.googleapis.com/auth/cloud-platform,https://www.googleapis.com/auth/pubsub"
       },
       "packageName": "google.pubsub.v1",
-      "iamService": false,
```
summer-ji-eng, line 498: curious why do we remove this line instead of setting `"iamService": true`

summer-ji-eng (approved):
Looks great. I test successfully on local two ways:
1. Under `gapic-generator-typescript` 
```
bazel run //:gapic_generator_typescript -- --output-dir /tmp/kms -I "$GOOGLEAPIS" --grpc-service-config "$GOOGLEAPIS/google/cloud/kms/v1/cloudkms_grpc_service_config.json" `find "$GOOGLEAPIS/google/cloud/kms/v1" -name '*.proto'` "$GOOGLEAPIS/google/cloud/common_resources.proto" --service_yaml="$GOOGLEAPIS/google/cloud/kms/v1/cloudkms.yaml"
```

2.  Under`$GOOGLEAPIS`
```
bazel build //google/cloud/kms/v1:kms-v1-nodejs
```
and remove the `iamService = true` ([line](https://github.com/googleapis/googleapis/blob/master/google/cloud/kms/v1/BUILD.bazel#L232
))

Thanks @alicejli hard working. 👋

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -28,7 +28,7 @@ import * as gax from 'google-gax';
 import {Callback, CallOptions, Descriptors, ClientOptions
 {%- if service.longRunning.length > 0 %}, LROperation{%- endif -%}
 {%- if service.paging.length > 0 %}, PaginationCallback, GaxCall{%- endif -%}
-{%- if service.iamService > 0 %}, IamClient, IamProtos{%- endif -%}
+{%- if (service.IAMPolicyMixin) > 0 %}, IamClient, IamProtos{%- endif -%}
```
alexander-fenster, line 31: Brackets are not needed (here and below).

test-fixtures/protos/google/showcase/v1beta1/showcase_v1beta1.yaml
```diff
@@ -0,0 +1,84 @@
...
+    - post: '/v1beta1/{resource=sequences/*}:testIamPermissions'
+      body: '*'
+  - selector: google.longrunning.Operations.ListOperations
+    get: '/v1beta1/operations'
+  - selector: google.longrunning.Operations.GetOperation
+    get: '/v1beta1/{name=/operations/*}'
+  - selector: google.longrunning.Operations.DeleteOperation
+    delete: '/v1beta1/{name=/operations/*}'
+  - selector: google.longrunning.Operations.CancelOperation
+    post: '/v1beta1/{name=/operations/*}:cancel'
```
alexander-fenster, line 84: Please add an empty line here :)

test-fixtures/protos/google/showcase/v1beta1/showcase_v1beta1.yaml
```diff
@@ -0,0 +1,84 @@
...
+    post: '/v1beta1/{resource=users/*}:setIamPolicy'
+    body: '*'
+    additional_bindings:
+    - post: '/v1beta1/{resource=rooms/*}:setIamPolicy'
+      body: '*'
+    - post: '/v1beta1/{resource=rooms/*/blurbs/*}:setIamPolicy'
+      body: '*'
+    - post: '/v1beta1/{resource=sequences/*}:setIamPolicy'
+      body: '*'
+  - selector: google.iam.v1.IAMPolicy.GetIamPolicy
```
alexander-fenster, line 61: I'm not sure if we want to add IAM methods to Showcase. Not sure if the server implementation really supports IAM. @noahdietz can you suggest here?  If Showcase server does not support IAM, we should instead have it covered by a different baseline test (e.g. KMS).

typescript/src/generator.ts
```diff
@@ -140,12 +143,24 @@ export class Generator {
...
+      const filename = this.paramMap['service-yaml'];
+      if (!fs.existsSync(filename)) {
+        throw new Error(`File ${filename} cannot be opened.`);
+      }
+      const content = fs.readFileSync(filename, 'utf8');
+      const info = yaml.load(content) as ServiceYaml;
+      this.serviceYaml = info;
+      const serviceMixins = [];
+      for (let i = 0; i < info.apis.length; i++) {
+        const api = JSON.stringify(info.apis[i]);
```
alexander-fenster, line 159: This is scary! Normally we don't need to go through `JSON.stringify` if we want to extract something from an object. Can you explain what you are doing here? Even if this is the right way to do it, a comment is definitely needed.

typescript/test/unit/baselines.ts
```diff
@@ -45,7 +45,7 @@ describe('Baseline tests', () => {
     outputDir: '.test-out-kms',
     protoPath: 'google/cloud/kms/v1/*.proto',
     useCommonProto: false,
-    iamService: true,
+    serviceYaml: 'google/cloud/kms/v1/cloudkms_v1.yaml',
```
alexander-fenster, line 48: The fact that the KMS baseline tests did not change basically means that your fixes are correct. There is no need to add IAM to Showcase.

alexander-fenster (changes requested):
A little bit concerned about the `JSON.stringify` thing, other than that, LGTM. Let's figure out what's going on there :)

typescript/src/generator.ts
alicejli: Sure! So ultimately I'm trying to create an array of strings, where each string is the name of an API. This array is what I'm using in the `proto.ts` file to derive whether or not the IAMPolicy mixin should be included or not. I wanted an array of strings for ease of comparison as it seems object comparison in Javascript is a little wonky. Is `JSON.stringify` specifically an issue, or should I re-evaluate the approach of creating an array of strings in general?

test-fixtures/protos/google/showcase/v1beta1/showcase_v1beta1.yaml
alicejli: Sorry - where exactly should the empty line go? I copied it from the master file: https://github.com/googleapis/gapic-showcase/blob/master/schema/google/showcase/v1beta1/showcase_v1beta1.yaml#81

alicejli: Ah never mind, realized it needs to be at the end. Thanks!

templates/typescript_gapic/src/$version/$service_client.ts.njk
alicejli: Thank you! Fixed.

typescript/src/generator.ts
alicejli: Edited to access the APIs directly from the object instead of using `JSON.stringify`. Thanks @summer-ji-eng for your help here.

test-fixtures/protos/google/showcase/v1beta1/showcase_v1beta1.yaml
alicejli: Confirmed with @noahdietz that the IAM Policy RPCs are supported in Showcase: https://github.com/googleapis/gapic-showcase/blob/master/server/services/iam_policy_service.go

So I kept the IAM methods in Showcase - give that it is covered in KMS though, let me know if you'd still prefer them to be removed from Showcase.

typescript/src/serviceyaml.ts
```diff
@@ -0,0 +1,18 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export interface ServiceYaml {
+  title: string;
+  apis: string[];
```
noahdietz, line 17: At some point, maybe not in this PR as it is rather focused, you will want the `http` rules as well.

noahdietz (commented):
Two things:
1. Does it make sense to keep the `iamService` flag for migration purposes? 
2. If the generator options are changing, the bazel rule should also change to reflect that

noahdietz:
> 2\. If the generator options are changing, the bazel rule should also change to reflect that

Never mind! It was already done, I just missed it

typescript/src/serviceyaml.ts
alicejli: Good to know! Created a quick issue for it here: https://github.com/googleapis/gapic-generator-typescript/issues/929 and also updated the `README` with the new `bazel run` command to include the `$service.yaml` file.

alicejli: Oh and chatted with Alex and Summer about the `iamService` flag - we decided that we'll go ahead with removing the `iamService` flag and bump the generator version (since it is technically a breaking change), and then update the `//third_party/googleapis/WORKSPACE`, `//third_party/googleapis_private/WORKSPACE`, and the `BUILD` files for `kms` and `pubsub` to use the latest generator version which should cover all the protos currently using `IAMPolicy`.

alexander-fenster (approved):

noahdietz (approved):
Well done Alice!

PR #916, opened 2021-06-30
feat: --legacy-proto-load to load proto files at runtime

alexander-fenster:
Implement a flag to enable "legacy proto loading": load `*.proto` files at runtime, without reading the huge compiled JSON. Should improve the runtime for Ads libraries.

Note: this option disables fallback mode and removes all mentions of the fallback.

Added one baseline test (regular showcase with this option enabled). No changes to the existing baseline tests (if an option is not given, existing libraries will not change).

Cc: @aohren


baselines/showcase-legacy/src/v1beta1/echo_client.ts.baseline
```diff
@@ -0,0 +1,818 @@
...
+    if (typeof process !== 'undefined' && 'versions' in process) {
+      clientHeader.push(`gl-node/${process.versions.node}`);
+    } else {
+      clientHeader.push(`gl-web/${this._gaxModule.version}`);
+    }
+    clientHeader.push(`grpc/${this._gaxGrpc.grpcVersion}`);
+    if (opts.libName && opts.libVersion) {
+      clientHeader.push(`${opts.libName}/${opts.libVersion}`);
+    }
+    // Load the applicable protos.
```
alexander-fenster, line 135: Note to reviewers: this is where the main change is.

summer-ji-eng (approved):
Thank @alexander-fenster working on this. 💯

PR #922, opened 2021-07-12
feat(diregapic)(pagination): support paging map response type

summer-ji-eng:
Before: paging response is only array.
After: support paging map type response for DIREGAPIC mode.

Map handle details:
gax-nodejs `createIterator` return array of tuple with pair of key and value. [Impl PR](https://github.com/googleapis/gax-nodejs/pull/1052).
The client library surface need to handle the return type as `AsyncIterable<[string, AddressAggregateList]>`

Next:
- Add compute_small.proto[https://github.com/googleapis/googleapis-discovery/blob/master/google/cloud/compute/v1/compute_small.proto] #930 
- add the test


templates/typescript_gapic/_util.njk
```diff
@@ -231,14 +231,22 @@ limitations under the License.
  * @returns {Object}
  *   An iterable Object that allows [async iteration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols).
  *   When you iterate the returned iterable, each element will be an object representing
+ {%- if method.pagingMapResponseType %}
+ *   {{ typeLink(method.pagingMapResponseType) }}. The API will be called under the hood as needed, once per the page,
```
alexander-fenster, line 235: This is not exactly correct: in this case each element will be an array of two elements, the first one is a string, and the second is pagingMapResponseType object, right?  The comment needs to be updated here.

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -683,6 +683,7 @@ export class {{ service.name }}Client {
   }
 {%- endfor %}
 {%- for method in service.paging %}
+  {%- if not method.pagingMapResponseType %}
```
alexander-fenster, line 686: nit: please align `if` and its corresponding `endif` (we tend not to offset nested template instructions to the right)

typescript/src/schema/proto.ts
```diff
@@ -44,6 +44,7 @@ interface MethodDescriptorProto
     | undefined;
   pagingFieldName: string | undefined;
   pagingResponseType?: string;
+  pagingMapResponseType?: string | undefined;
```
alexander-fenster, line 47: No need to have `| undefined` since the question mark does just that.

typescript/src/schema/proto.ts
```diff
@@ -282,6 +283,50 @@ function pagingResponseType(
...
 
+// Support DIREGAPIC google-discovery API pagination response with map field.
+function pagingMapResponseType(
+  messages: MessagesMap,
+  method: MethodDescriptorProto,
+  rest?: boolean
+) {
+  const pagingfield = pagingField(messages, method, undefined, rest);
+  const outputType = messages[method.outputType!];
+  if (!pagingfield || !pagingfield.type || !rest || !outputType.nestedType) {
```
alexander-fenster, line 294: `!pagingfield || !pagingfield.type` can probably be replaced with `!pagingfield?.type`  (the `?.` operator will return `undefined` if `pagingfield` is `undefined`)

typescript/src/schema/proto.ts
```diff
@@ -282,6 +283,50 @@ function pagingResponseType(
...
+  if (mapResponses.length === 0) {
+    return undefined;
+  }
+  if (mapResponses.length > 1) {
+    throw new Error(
+      `Paginated "${method.name}" method can only have one map field`
+    );
+  }
+  const pagingMapResponse = mapResponses[0];
+  if (pagingMapResponse.name !== pagingfield.typeName?.split('.').pop()) {
```
alexander-fenster, line 309: Can you explain what are you checking here?  Maybe a comment would be useful here.

alexander-fenster (approved):
Looks good so far! Would love to see how this would change the compute_small baseline. Let's add the baseline first, and then update this PR.

templates/typescript_gapic/_util.njk
summer-ji-eng: Agree. Updated.

templates/typescript_gapic/src/$version/$service_client.ts.njk
summer-ji-eng: sg

typescript/src/schema/proto.ts
summer-ji-eng: 👍

typescript/src/schema/proto.ts
summer-ji-eng: 👍 thanks for the explanation.

typescript/src/schema/proto.ts
summer-ji-eng: // convert paging.field typeName .google.cloud.compute.v1.ItemEntry to ItemEntry
leave the comments.

summer-ji-eng:
Thanks for reviewing my PRs. 🙇  Baseline PR is here #930

PR #930, opened 2021-07-14
feat(diregapic): add compute proto in baseline

summer-ji-eng:
Add baseline test for compute by adding [compute_small.proto](https://github.com/googleapis/googleapis-discovery/blob/master/google/cloud/compute/v1/compute_small.proto)

With this latest generated compute client library, it support the pagination for `maxResults` and handle the map field in paging response.


alexander-fenster:
Since compute is totally different library, it makes sense also to have it in GitHub Actions. Can you also add it here:

https://github.com/googleapis/gapic-generator-typescript/blob/master/.github/workflows/ci.yaml#L117

that way the CI will compile, test, and pack the generated compute library and we'll see any compilation errors, if any. Thank you!

alexander-fenster (approved):
Approved with 1 comment, thank you!

summer-ji-eng:
fix unit test in #934

PR #936, opened 2021-07-15
fix(legacy proto load): use single proto file

alexander-fenster:
Also, removes proto JSON file if legacy proto options is enabled since it's not used but consumes space and takes time to generate.


alexander-fenster:
Cc: @aohren

aohren:
Tested and it looks like this works. And there are some noticeable performance improvements.

summer-ji-eng (approved):

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -253,11 +247,9 @@ export class {{ service.name }}Client {
 {%- if api.legacyProtoLoad %}
 
     let protoFilesRoot = new this._gaxModule.GoogleProtoFilesRoot();
-{%- for filename in api.filesToGenerate %}
```
summer-ji-eng, line 256: One question, why do we need to remove the filesToGenerate condition? We only has one proto to generate?

alexander-fenster: Yes, the idea is to have just one proto here (the one that defines the service).  If we find it does not work for some use cases, we'll think how to fix it.

PR #942, opened 2021-07-28
feat: adding operations mixin service

alicejli:
For googleapis/google-cloud-node-core#610 

As part of adding support generally for client library mixin method generations (draft AIP is at aip-dev/google.aip.dev#694), this update is to update the generator to add in google.longrunning.Operations service as a mixin for any api that includes it in their <$service>.yaml file.

Ping me for the internal doc(s) if desired.

### DIFF
--- a/baselines/showcase/src/v1beta1/echo_client.ts.baseline
+++ b/baselines/showcase/src/v1beta1/echo_client.ts.baseline
@@ -986,6 +986,172 @@ export class EchoClient {
     return this.iamClient.testIamPermissions(request, options, callback);
   }
 
+/**
+   * Gets the latest state of a long-running operation.  Clients can use this
+   * method to poll the operation result at intervals as recommended by the API
+   * service.
+   *
+   * @param {Object} request - The request object that will be sent.
+   * @param {string} request.name - The name of the operation resource.
+   * @param {Object=} options
+   *   Optional parameters. You can override the default settings for this call,
+   *   e.g, timeout, retries, paginations, etc. See [gax.CallOptions]{@link
+   *   https://googleapis.github.io/gax-nodejs/global.html#CallOptions} for the
+   *   details.
+   * @param {function(?Error, ?Object)=} callback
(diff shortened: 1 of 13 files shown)


baselines/showcase/src/v1beta1/echo_client.ts.baseline
```diff
@@ -986,6 +986,172 @@ export class EchoClient {
...
+    callback?: Callback<
+      protos.google.longrunning.Operation,
+      protos.google.longrunning.GetOperationRequest,
+      {} | null | undefined
+    >
+  ): Promise<[protos.google.longrunning.Operation]> {
+    return this.operationsClient.getOperation(request, options, callback);
+  }
+  /**
+   * Equivalent to {@link listOperations}, but returns an iterable object.
```
alexander-fenster, line 1038: ...but there's no `listOperations` :)  We need to update the comment here.

alexander-fenster: Also, can we have an in-code `@example` for this too? 

```
for await (const response of client.listOperationsAsync(request)) {
  // doThingsWith(response);
}
```

alicejli: good catch! thanks - updated comment and example.

typescript/src/schema/proto.ts
```diff
@@ -572,6 +573,13 @@ function augmentService(parameters: AugmentServiceParameters) {
   ) {
     augmentedService.IAMPolicyMixin = 1;
   }
+  if (
+    parameters.options.serviceYaml?.apis.includes(
+      'google.longrunning.Operations'
+    )
+  ) {
+    augmentedService.LongRunningOperationsMixin = 1;
```
summer-ji-eng, line 581: I think we should make it boolean for better readability.

summer-ji-eng: Feel free to do that in separate PR.

summer-ji-eng (approved):

PR #953, opened 2021-08-11
feat: close mixin clients

alicejli:
For services that use a mixin client (IAMPolicy, Locations, LongRunningOperations), when they call `close()` on the client, it does not currently close the mixin client. If one is generating many clients in a row, this can cause memory leaks (see: https://github.com/googleapis/google-cloud-node-core/issues/625).

This feature closes the mixin clients in addition to the service client when the end user calls `close()`. 

This significantly improves memory leakage for clients using LongRunningOperations such as `InstanceAdminClient`.


summer-ji-eng (approved):

summer-ji-eng:
💯 wow, thank you debugging on this.

PR #956, opened 2021-08-17
feat: rename unary LRO for diregapic

summer-ji-eng:
Rename DIREGAPIC LRO for forward compatibility at GA

### DIFF
--- a/baselines/compute/src/v1/addresses_client.ts.baseline
+++ b/baselines/compute/src/v1/addresses_client.ts.baseline
@@ -206,7 +206,7 @@ export class AddressesClient {
     // Iterate over each of the methods that the service provides
     // and create an API call method for each.
     const addressesStubMethods =
-        ['aggregatedList', 'delete', 'insert', 'list'];
+        ['aggregatedList', 'unaryDelete', 'unaryInsert', 'list'];
     for (const methodName of addressesStubMethods) {
       const callPromise = this.addressesStub.then(
         stub => (...args: Array<{}>) => {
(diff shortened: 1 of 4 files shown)


baselines/compute/src/v1/addresses_client_config.json.baseline
```diff
@@ -24,11 +24,11 @@
           "retry_codes_name": "non_idempotent",
           "retry_params_name": "default"
         },
-        "Delete": {
+        "UnaryDelete": {
```
alexander-fenster, line 27: I believe we must keep the original method names in this config because this is only a client change and gax will keep using original names.

baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -206,7 +206,7 @@ export class AddressesClient {
     // Iterate over each of the methods that the service provides
     // and create an API call method for each.
     const addressesStubMethods =
-        ['aggregatedList', 'delete', 'insert', 'list'];
+        ['aggregatedList', 'unaryDelete', 'unaryInsert', 'list'];
```
alexander-fenster, line 209: We must not rename the names in the stubs. The stub still has original names taken from the protos.

baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -369,23 +369,23 @@ export class AddressesClient {
       'project': request.project || '',
     });
     this.initialize();
-    return this.innerApiCalls.delete(request, options, callback);
+    return this.innerApiCalls.unaryDelete(request, options, callback);
```
alexander-fenster, line 372: same here, let's keep `innerApiCalls` name just `delete`. We only want to rename the user-visible method, without touching any internals.

typescript/src/schema/proto.ts
```diff
@@ -438,6 +458,11 @@ function augmentMethod(
       );
     }
   }
+  // Rename the unary LRO method
+  if (isUnaryLRO(parameters.service, method, parameters.diregapic)) {
+    method.name = 'Unary' + method.name;
```
alexander-fenster, line 463: This is not the right approach. The generator code must not know anything about the TypeScript specific. The renames must be done in the templates, there is a Namer functionality there to avoid naming conflicts.

See how we build the names for paginated Async function as an example.

vchudnov-g:
Thanks for doing this! Two notes:
- let's not merge this until we have agreement that providing the unary methods (as opposed to, say, just doing the full implementation outright) is what we want to do
- if we do decide to go with this approach, it would be good for the name modification to be as consistent as possible across languages. See the internal doc for the possibilities we're considering. I'm guessing it will be a suffix, not a prefix, and something other than "Unary"

summer-ji-eng:
Another approach on #967

PR #967, opened 2021-08-30
feat(diregapic)(LRO): Add minimum wrapper for mitigate future full-implementation diregapic LRO change

summer-ji-eng:
Wrap an unary operations, that made it return a response compatible with the [eventual LRO operation](https://github.com/googleapis/gax-nodejs/blob/master/src/clientInterface.ts#L62). In another word, utilize duck typing to return the similar structure, but no ".promise" field.

Sample code test on generated client library:
```
async function main() {
  const projectId = await regionOperationClient.getProjectId();
  const region = 'us-central1';
  const rand = Math.floor(Math.random() * 1000);
  const addressResource = {
    name: `operation-insert-address-test-${rand}`,
  };
  const [operation] = await addressClient.insert({
    addressResource: addressResource,
    project: projectId,
    region: region
  })
  console.log('insert address operation.id:: ', operation);
  }
```
The output:
insert address operation.id::  {
  latestResponse: {
    warnings: [],
    id: '3894509470566436984',
    _id: 'id',
    startTime: '2021-08-29T22:37:27.644-07:00',
    _startTime: 'startTime',
    targetLink: 'https://www.googleapis.com/compute/v1/projects/gapic-images/regions/us-central1/addresses/operation-insert-address-test-188',
    _targetLink: 'targetLink',
    progress: 0,
    _progress: 'progress',
    region: 'https://www.googleapis.com/compute/v1/projects/gapic-images/regions/us-central1',
    _region: 'region',
    operationType: 'insert',
    _operationType: 'operationType',
    status: 'RUNNING',
    _status: 'status',
    ...
  },
  done: false,
  name: '3894509470566436984',
  metadata: null,
  result: {}
}

### DIFF
--- a/baselines/compute/src/v1/addresses_client.ts.baseline
+++ b/baselines/compute/src/v1/addresses_client.ts.baseline
@@ -18,7 +18,7 @@
 
 /* global window */
 import * as gax from 'google-gax';
-import {Callback, CallOptions, Descriptors, ClientOptions, PaginationCallback, GaxCall} from 'google-gax';
+import {Callback, CallOptions, Descriptors, ClientOptions, LROperation, PaginationCallback, GaxCall} from 'google-gax';
 
 import { Transform } from 'stream';
 import { RequestType } from 'google-gax/build/src/apitypes';
@@ -297,8 +297,8 @@ export class AddressesClient {
       request?: protos.google.cloud.compute.v1.IDeleteAddressRequest,
       options?: CallOptions):
       Promise<[
-        protos.google.cloud.compute.v1.IOperation,
-        protos.google.cloud.compute.v1.IDeleteAddressRequest|undefined, {}|undefined
+        LROperation<protos.google.cloud.compute.v1.IOperation, null>,
+        protos.google.cloud.compute.v1.IOperation|undefined, {}|undefined
       ]>;
   delete(
       request: protos.google.cloud.compute.v1.IDeleteAddressRequest,
@@ -351,8 +351,8 @@ export class AddressesClient {
           protos.google.cloud.compute.v1.IDeleteAddressRequest|null|undefined,
           {}|null|undefined>):
       Promise<[
-        protos.google.cloud.compute.v1.IOperation,
-        protos.google.cloud.compute.v1.IDeleteAddressRequest|undefined, {}|undefined
+        LROperation<protos.google.cloud.compute.v1.IOperation, null>,
+        protos.google.cloud.compute.v1.IOperation|undefined, {}|undefined
       ]>|void {
     request = request || {};
     let options: CallOptions;
@@ -372,14 +372,33 @@ export class AddressesClient {
       'project': request.project || '',
     });
     this.initialize();
-    return this.innerApiCalls.delete(request, options, callback);
+    return new Promise<[
+        LROperation<protos.google.cloud.compute.v1.IOperation, null>,
+        protos.google.cloud.compute.v1.IOperation|undefined, {}|undefined
+      ]>((resolve, reject) => {
+      this.innerApiCalls.delete(request, options, callback)
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      .then(([response, next, rawResponse]: [any, any, any]) => {
+        resolve([
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-expect-error
+          { latestResponse: response, done: false, name: response.id},
+          next,
+          rawResponse
+        ]);
+      })
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      .catch((err: any) => {
+        reject(err);
+      })
+    });
   }
   insert(
       request?: protos.google.cloud.compute.v1.IInsertAddressRequest,
       options?: CallOptions):
       Promise<[
-        protos.google.cloud.compute.v1.IOperation,
-        protos.google.cloud.compute.v1.IInsertAddressRequest|undefined, {}|undefined
+        LROperation<protos.google.cloud.compute.v1.IOperation, null>,
+        protos.google.cloud.compute.v1.IOperation|undefined, {}|undefined
       ]>;
   insert(
       request: protos.google.cloud.compute.v1.IInsertAddressRequest,
@@ -432,8 +451,8 @@ export class AddressesClient {
           protos.google.cloud.compute.v1.IInsertAddressRequest|null|undefined,
           {}|null|undefined>):
       Promise<[
-        protos.google.cloud.compute.v1.IOperation,
-        protos.google.cloud.compute.v1.IInsertAddressRequest|undefined, {}|undefined
+        LROperation<protos.google.cloud.compute.v1.IOperation, null>,
+        protos.google.cloud.compute.v1.IOperation|undefined, {}|undefined
       ]>|void {
     request = request || {};
     let options: CallOptions;
@@ -453,7 +472,26 @@ export class AddressesClient {
       'project': request.project || '',
     });
     this.initialize();
-    return this.innerApiCalls.insert(request, options, callback);
+    return new Promise<[
+        LROperation<protos.google.cloud.compute.v1.IOperation, null>,
+        protos.google.cloud.compute.v1.IOperation|undefined, {}|undefined
+      ]>((resolve, reject) => {
+      this.innerApiCalls.insert(request, options, callback)
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      .then(([response, next, rawResponse]: [any, any, any]) => {
+        resolve([
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-expect-error
+          { latestResponse: response, done: false, name: response.id},
+          next,
+          rawResponse
+        ]);
+      })
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      .catch((err: any) => {
+        reject(err);
+      })
+    });
   }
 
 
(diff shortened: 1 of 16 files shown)


baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -372,14 +372,33 @@ export class AddressesClient {
...
+    return new Promise<[
+        LROperation<protos.google.cloud.compute.v1.IOperation, null>,
+        protos.google.cloud.compute.v1.IOperation|undefined, {}|undefined
+      ]>((resolve, reject) => {
+      this.innerApiCalls.delete(request, options, callback)
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      .then(([response, next, rawResponse]: [any, any, any]) => {
+        resolve([
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-expect-error
```
summer-ji-eng, line 385: I have question to use duck typing for interface [gax.Operation](https://github.com/googleapis/gax-nodejs/blob/acedd552199f57ffae23ed0c41464304fc3bb6f2/src/longRunningCalls/longrunning.ts#L46). The current object has only 3 fields: `latestResponse, done, name`, ts linter compliant that miss other fields in [Operation](https://github.com/googleapis/gax-nodejs/blob/acedd552199f57ffae23ed0c41464304fc3bb6f2/src/longRunningCalls/longrunning.ts#L46).
Option 1(fix the linter):
If I try to pass non-optional value for the rest fields, there are two fields I am not sure how to sign value `longrunningDescriptor` and `backoffSettings`. We need to construct these two objects but eventually will replace after full-implementation.

Option2(ignore linter)
Another option is what I did here, ignore the linter. And save some time for full implementation and remove entire wrapper as soon as we fully implement diregapic lro.

@alexander-fenster, @bcoe can I have some advice here. Or maybe we have easy quick way to fix.

alexander-fenster: You can put a fake LongRunningDescriptor (just create a new one) and a fake backoff settings (again, there should be a way to create a default backoff settings object). Replacing them with real stuff later won't break compatibility.

summer-ji-eng: Oh, sure. Thanks for clarification. I will go ahead to return the response same as LROperation.

One reason I hesitate to create a fake LongRunningDescriptor is because it surface the fake object to user in the generated client library. I don't have a good reason say that we should not do this. Just not super sure. Now I got more confidence to do so.

baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -297,8 +297,8 @@ export class AddressesClient {
       request?: protos.google.cloud.compute.v1.IDeleteAddressRequest,
       options?: CallOptions):
       Promise<[
-        protos.google.cloud.compute.v1.IOperation,
-        protos.google.cloud.compute.v1.IDeleteAddressRequest|undefined, {}|undefined
+        LROperation<protos.google.cloud.compute.v1.IOperation, null>,
```
alexander-fenster, line 300: is the metadata going to be `null` when we implement real LRO?

baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -333,12 +333,13 @@ export class AddressesClient {
  * @param {object} [options]
  *   Call options. See {@link https://googleapis.dev/nodejs/google-gax/latest/interfaces/CallOptions.html|CallOptions} for more details.
  * @returns {Promise} - The promise which resolves to an array.
- *   The first element of the array is an object representing [Operation]{@link google.cloud.compute.v1.Operation}.
+ *   The first element of the array is an object representing
+ *   a long running operation.
  *   Please see the
- *   [documentation](https://github.com/googleapis/gax-nodejs/blob/master/client-libraries.md#regular-methods)
+ *   [documentation](https://github.com/googleapis/gax-nodejs/blob/master/client-libraries.md#long-running-operations)
```
alexander-fenster, line 339: Let's say here that awaiting for `.promise()` is not supported yet (in "if diregapic" block)

baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -372,14 +373,31 @@ export class AddressesClient {
       'project': request.project || '',
     });
     this.initialize();
-    return this.innerApiCalls.delete(request, options, callback);
+    return new Promise<[
```
alexander-fenster, line 376: Why do you need to create a new Promise here? Can't we just do `return this.innerApiCalls.delete(...).then(...)`?  Also, there will be no need to `.catch` anything since the error will be just thrown and the promise will get rejected.

baselines/compute/src/v1/addresses_client.ts.baseline
summer-ji-eng: Add comment for beta. Open to any suggestion.

baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -333,12 +333,17 @@ export class AddressesClient {
...
+ *   The first element of the array is an object representing
+ *   a long running operation.
  *   Please see the
- *   [documentation](https://github.com/googleapis/gax-nodejs/blob/master/client-libraries.md#regular-methods)
+ *   [documentation](https://github.com/googleapis/gax-nodejs/blob/master/client-libraries.md#long-running-operations)
  *   for more details and examples.
+ *   This method is considered to be in beta. This means while
+ *   stable it is still a work-in-progress and under active development,
+ *   and might get backwards-incompatible changes at any time.
+ *   `.promise()` is not supported yet.
```
summer-ji-eng, line 344: @bcoe any feedback or suggestion on comment tell user we are not implement `.promise` yet. ✋

baselines/compute/src/v1/addresses_client.ts.baseline
summer-ji-eng: I think so. MetadataType contains the status of the operation. In compute, it contains the similar status field in [operation](https://github.com/googleapis/googleapis-discovery/blob/master/google/cloud/compute/v1/compute_small.proto#L157). As that being said, we are not support `MetadataType` in diregapic LRO.
(but please correct me if I misunderstanding here) 🙏

baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -453,7 +471,15 @@ export class AddressesClient {
       'project': request.project || '',
     });
     this.initialize();
-    return this.innerApiCalls.insert(request, options, callback);
+    return this.innerApiCalls.insert(request, options, callback)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .then(([response, next, rawResponse]: [any, any, any]) => {
+      return [
```
bcoe, line 477: This is exactly what I was picturing, I'm glad it wasn't too painful to make the data fit the appropriate shape.

bcoe (approved):

typescript/src/schema/proto.ts
```diff
@@ -159,6 +161,26 @@ function longRunningMetadataType(
...
+  isDiregapic?: boolean
+): boolean {
+  const operationOutputType = toFullyQualifiedName(packageName, 'Operation');
+  return isDiregapic &&
+    method.outputType &&
+    method.outputType === operationOutputType &&
+    method.options?.['.google.api.http'] &&
+    !(
+      method.options?.['.google.api.http'].get ||
+      (method.name === 'Wait' && method.options?.['.google.api.http'].post)
```
bcoe, line 178: Are we saying that any POST method will be long running, unless it's of type `Wait`?

baselines/compute/src/v1/addresses_client.ts.baseline
summer-ji-eng: synced offline

typescript/src/schema/proto.ts
summer-ji-eng: Yes

typescript/src/schema/proto.ts
```diff
@@ -159,6 +161,26 @@ function longRunningMetadataType(
...
+    )
+    ? true
```
alexander-fenster, line 180: nit: there is no need to write `booleanValue ? true : false`, just `booleanValue` is enough.  So you can just remove lines 180 and 181.

baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -372,14 +377,22 @@ export class AddressesClient {
       'project': request.project || '',
     });
     this.initialize();
-    return this.innerApiCalls.delete(request, options, callback);
+    return this.innerApiCalls.delete(request, options, callback)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    .then(([response, next, rawResponse]: [any, any, any]) => {
```
alexander-fenster, line 382: It's not `next` here (it's not the pagination).  The right names would be `[response, operation, rawResponse]`. Also, we know exact types: all three have the same type `protos.google.cloud.compute.v1.IOperation` in our case :)  So we don't need to use `any` here.

alexander-fenster (changes requested):
Looks good to me, let's just get rid of `any` :)

baselines/compute/src/v1/addresses_client.ts.baseline
summer-ji-eng: sg

typescript/src/schema/proto.ts
summer-ji-eng: I removed line 180 and 181, but update `isDiregapic` with the type `boolean | '' | null | undefined`

baselines/compute/src/v1/addresses_client.ts.baseline
```diff
@@ -372,14 +377,22 @@ export class AddressesClient {
       'project': request.project || '',
     });
     this.initialize();
-    return this.innerApiCalls.delete(request, options, callback);
+    return this.innerApiCalls.delete(request, options, callback)
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
alexander-fenster, line 381: nit: no need to disable `no-explicit-any` now :)

alexander-fenster (approved):

PR #973, opened 2021-09-03
feat: update retry and timeout logic for generated async/stream methods

alicejli:
This updates generated Stream and Async methods to use the default retry configuration of the base method if set by the service.
For instance, with this fix, BaseMethodStream and BaseMethodAsync will use the retry configuration of the BaseMethod method. Originally, they were using a default timeout of 30s with no retry configuration.

Please ping me for internal docs if desired.


summer-ji-eng (approved):
Confirmed that we only have pagination method Stream/Async are generated. 👍

PR #1007, opened 2021-10-07
feat: adds dynamic routing header functionality

alicejli:
For issue: https://github.com/googleapis/google-cloud-node-core/issues/584

For the purpose of routing at GFE level, there is a need for a part of RPC’s input message to be extracted and put into a header. This is called ‘dynamic routing’ and the header ‘the routing header’. There are 3 teams that currently have a need for this functionality: BigTable, Spanner and Firestore.

Currently there is a support for dynamic routing in client libraries – the code fills the x-goog-request-params header with the string constructed from the values of the fields bound to the URI in the google.api.http annotation. However, several teams want the ability to specify a more specific header. This header will be specified in a new google.api.routing annotation, detailed as such:

// Specification of information to add to facilitate correct routing of the API requests.
message RoutingRule {
  repeated HeaderParameter routing_parameters = 1;
}
where HeaderParameter is

// A projection from an input message to the routing header
message HeaderParameter {
  // A field to extract the header key-value pair from
  string field = 1;
  // A pattern matching the key-value field. Optional.
  // If not specified, the whole field specified in the `field` field will be taken as value, and its name used as key.
  string path_template = 2;
}

TODO: The ability to override this annotation will exist in the ${service}.yaml file for each API. Once that is built, will need to add in that as a source for where the generator checks for the headers to generate.

Please ping me for additional internal docs or info if needed.


typescript/src/util.ts
```diff
@@ -179,3 +179,160 @@ export function getResourceNameByPattern(pattern: string): string {
...
+
+// This intakes a path template and returns an array where the first element is the full named
+// segment, the second element is the name of the segment, the third element is the segment itself,
+// and the fourth element is the named capture regex of the named segement.
+export function getNamedSegment(pattern: string): string[] {
+  //Named segment in path template will always be contained within '{}'
+  const curlyBraceRegex = new RegExp(/[^{}]+(?=})/g);
+  const element = pattern.match(curlyBraceRegex);
+  if (element) {
+    if (checkIfArrayContainsOnlyOneNamedSegment(element!)) {
```
summer-ji-eng, line 304: nit: it should safe to remove `!`, because we checked `elememt` exist in line 303

typescript/src/util.ts
```diff
@@ -179,3 +179,160 @@ export function getResourceNameByPattern(pattern: string): string {
...
+      return namedSegment;
+    } else {
+      throw new Error(
+        'Path template must contain exactly one named segment - e.g. {foo=bar}.'
+      );
+    }
+  } else {
+    throw new Error(
+      'Path template must contain exactly one named segment - e.g. {foo=bar}.'
+    );
```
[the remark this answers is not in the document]
summer-ji-eng: beside throw Error, can we return empty array instead.

typescript/src/schema/proto.ts
```diff
@@ -59,6 +59,13 @@ interface MethodDescriptorProto
   // into x-goog-request-params header, the array will contain
   // [ ['request', 'foo'], ['request', 'bar']]
   headerRequestParams: string[][];
+  // if 1; send implicit header request. If 2, send dynamic header request. If 3, do not send a header at all.
+  headerRequest: number;
```
summer-ji-eng, line 63: For better readability, could we use boolean to indicate the dynamic header. See below for sample refactor code.

typescript/src/schema/proto.ts
```diff
@@ -539,9 +534,33 @@ function augmentMethod(
...
+    // if dynamic routing rules are empty, then do not set a header
+    if (
+      method.dynamicHeaderRequestParams === null ||
+      method.dynamicHeaderRequestParams.length < 1
+    ) {
+      method.headerRequest = 3;
+    } else {
+      method.headerRequest = 2;
+    }
+  }
```
summer-ji-eng, line 563: If we use boolean to indicate the dynamic routing, we could simplify similar as 
```
method.headerRequestParams = getHeaderRequestParams(
      method.options?.['.google.api.http']
    );
if (methodDynamicRouting && methodDynamicRoutingParameters && methodDynamicRoutingParameters.length > 1) {
	isDynamicHeader = true;
	method.dynamicHeaderRequestParams = getDynamicHeaderRequestParams(
      methodDynamicRoutingParameters!
    );
}
```

Is that possible that a method has both `.google.api.http` and `.google.api.routing` annotation?

typescript/src/schema/proto.ts
```diff
@@ -576,10 +595,120 @@ export function getHeaderRequestParams(
...
+) {
+  const params: string[][][] = [[]];
+  let counterOfParameters = 0;
+  let counterOfRules = 0;
+  const setOfParameters = new Set();
+  // if there are no rules or the rules are empty or there is a malformed routing rule, return empty array
+  try {
+    if (
+      rules === null ||
+      rules.length === 0 ||
```
summer-ji-eng, line 627: we could remove check for null case, if the useage is similar to 
```
if (methodDynamicRouting && methodDynamicRoutingParameters && methodDynamicRoutingParameters.length > 1) {
	method.dynamicHeaderRequestParams = getDynamicHeaderRequestParams(
      methodDynamicRoutingParameters!
    );
}
```

typescript/src/schema/proto.ts
```diff
@@ -576,10 +595,120 @@ export function getHeaderRequestParams(
...
+}
+
+// This parses a single Routing Parameter and returns an array with the first element being the name of the
+// field the header should retrieve from the message, the second element being the name of the field the header should send,
+// the third element is the full regex of the path template that the message field should match,
+// and the last element is the named regex capture of the value of the field.
+export function getSingleRoutingHeaderParam(
+  rule: protos.google.api.IRoutingParameter
+) {
+  const mapOfRoutingParameters: string[] = [];
```
summer-ji-eng, line 694: For better readability, we could data like `map`, `Tuple`, `Interface` to store the data instead of use array[0], array[1], array[2]...
Use Tuple example:
```
interface MapRoutingParam {
 field: string,
 nameSegment: string,
....
}
```

summer-ji-eng (changes requested):
Thank you @alicejli work on this feature.

After address the readability comments, could you ask @viacheslav-rostovtsev take another look on routing business.

summer-ji-eng (changes requested):
I would suggest you to split the PR into at least 2 PRs if you have bandwidth. First PR cover the util function and test. And rest logic for second PR.

alicejli:
I have split this PR into two new separate PRs: https://github.com/googleapis/gapic-generator-typescript/pull/1028 and https://github.com/googleapis/gapic-generator-typescript/pull/1042. Closing this one.

PR #1028, opened 2021-10-26
feat: add dynamic routing annotation parsing functionality

alicejli:
Breaking up https://github.com/googleapis/gapic-generator-typescript/pull/1007 into 2 different PRs for ease of review. This part adds in the ability to parse a dynamic routing header annotation and convert it into regex. 

Background:

For issue: googleapis/google-cloud-node-core#584

For the purpose of routing at GFE level, there is a need for a part of RPC’s input message to be extracted and put into a header. This is called ‘dynamic routing’ and the header ‘the routing header’. There are 3 teams that currently have a need for this functionality: BigTable, Spanner and Firestore.

Currently there is a support for dynamic routing in client libraries – the code fills the x-goog-request-params header with the string constructed from the values of the fields bound to the URI in the google.api.http annotation. However, several teams want the ability to specify a more specific header. This header will be specified in a new google.api.routing annotation, detailed as such:

```
// Specification of information to add to facilitate correct routing of the API requests.
message RoutingRule {
  repeated HeaderParameter routing_parameters = 1;
}
```
where HeaderParameter is

```
// A projection from an input message to the routing header
message HeaderParameter {
  // A field to extract the header key-value pair from
  string field = 1;
  // A pattern matching the key-value field. Optional.
  // If not specified, the whole field specified in the field field will be taken as value, and its name used as key.  
  string path_template = 2;
}
```

TODO: The ability to override this annotation will exist in the ${service}.yaml file for each API. Once that is built, will need to add in that as a source for where the generator checks for the headers to generate.

Please ping me for additional internal docs or info if needed.

### DIFF
--- a/typescript/src/util.ts
+++ b/typescript/src/util.ts
@@ -205,20 +205,16 @@ export function checkIfArrayContainsOnlyOneNamedSegment(
   pattern: string[]
 ): boolean {
   let countOfNamedSegments = 0;
-  let counter = 0;
   // Checks that resources are named
   const regexNamed = new RegExp(/^.+=.+/);
   const regexWildcard = new RegExp(/^{[a-zA-Z\d-]+}/);
-  pattern.forEach((element) => {
-    if (
-      regexNamed.test(element) ||
-      regexWildcard.test(element)
-    ) {
+  pattern.forEach(element => {
+    if (regexNamed.test(element) || regexWildcard.test(element)) {
       countOfNamedSegments++;
     }
-  })
-    return countOfNamedSegments === 1;
-  }
+  });
+  return countOfNamedSegments === 1;
+}
 
 // This takes in a path template segment and converts it into regex.
 export function convertSegmentToRegex(pattern: string): string {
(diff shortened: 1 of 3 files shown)


typescript/src/util.ts
```diff
@@ -190,3 +190,153 @@ export function getResourceNameByPattern(pattern: string): string {
...
+      regexWildcard.test(pattern[counter])
+    ) {
+      countOfNamedSegments++;
+    }
+    counter++;
+    if (countOfNamedSegments > 1) {
+      return false;
+    }
+  }
+  return countOfNamedSegments === 1;
```
summer-ji-eng, line 224: emm... nothing wrong, it's not block, but we could utilize Array.protoype.forEach()
`forEach((element, index) => { ... } )` for better readability and save one variable.

```
pattern.forEach((element) => {
   if (regexNamed.test(element) || regexWildcard.test(element)) {
       countOfNamedSegments++;
    }
})
return countOfNamedSegments === 1;

summer-ji-eng (approved):
Left a comment about the style, other than that looks good to me. Thanks @alicejli for your hard work.

typescript/src/util.ts
alicejli: Thank you! That is much cleaner - will take your suggestion.

PR #1030, opened 2021-10-27
feat: move comments in front of overloaded function

alicejli:
Google is moving reference docs to cloud.google.com, which relies on TSDoc rather than JSDoc.

TSDoc expects comments before the first overloaded function, we currently have those on the last function. Those comments don't make it into the d.ts files. We need to move comments to the first overloaded function rather than the last one.

This update moves the comments ahead of the first function instead of the last one with the same name.

See for example https://github.com/googleapis/nodejs-secret-manager/commit/980b0fdfd48b5282d450de16af26daef80f53e70


fhinkel (approved):
This looks great, thank you!

summer-ji-eng (approved):
Thank you @alicejli 💯

PR #1034, opened 2021-10-29
fix: naming collision on samples

### LINKED ISSUE #583: bug(sample-gen): do not use reserved words
https://github.com/googleapis/nodejs-functions/pulls

```
 then two changes are needed

(1) use the logic similar to the one in namer.js to replace the names if they are bad

(namer.js is basically a way to plug the JS code to the template execution phase)

and then (2) make sure the usages use the new name, so const request = { function: renamedFunction }

namer: https://github.com/googleapis/gapic-generator-typescript/blob/main/templates/typescript_gapic/namer.js

it can be preinitialized with things like function

usage: https://github.com/googleapis/gapic-generator-typescript/blob/main/templates/typescript_gapic/src/%24version/%24service_client.ts.njk#L823

(use id.get(suggestedName) everywhere where it might cause conflict, and it will be renamed by the namer if needed)


otherwise, in cases where a proto has a real naming conflict, the samples will use the wrong names

here, https://github.com/googleapis/gapic-generator-typescript/blob/main/templates/typescript_samples/samples/generated/%24version/%24service.%24method.js.njk#L48

the template says  async function {{ method.name.toCamelCase() }}() { while it should say async function {{ id.get(method.name.toCamelCase()) }} {

the rule is, all class members' names in the output should go through the namer

(and now we're going to expand this to the parameter names as well)

We need to make the namer consider reserved words like function as "used"

the logic is really straightforward, the only complicated thing here is how this logic is plugged in to the templates (which, as you surely know by now, don't allow much to do )
```

### LINKED ISSUE #1000: bug(sample-gen): do not use reserved words
https://github.com/googleapis/nodejs-functions/pulls

```
 then two changes are needed

(1) use the logic similar to the one in namer.js to replace the names if they are bad

(namer.js is basically a way to plug the JS code to the template execution phase)

and then (2) make sure the usages use the new name, so const request = { function: renamedFunction }

namer: https://github.com/googleapis/gapic-generator-typescript/blob/main/templates/typescript_gapic/namer.js

it can be preinitialized with things like function

usage: https://github.com/googleapis/gapic-generator-typescript/blob/main/templates/typescript_gapic/src/%24version/%24service_client.ts.njk#L823

(use id.get(suggestedName) everywhere where it might cause conflict, and it will be renamed by the namer if needed)


otherwise, in cases where a proto has a real naming conflict, the samples will use the wrong names

here, https://github.com/googleapis/gapic-generator-typescript/blob/main/templates/typescript_samples/samples/generated/%24version/%24service.%24method.js.njk#L48

the template says  async function {{ method.name.toCamelCase() }}() { while it should say async function {{ id.get(method.name.toCamelCase()) }} {

the rule is, all class members' names in the output should go through the namer

(and now we're going to expand this to the parameter names as well)

We need to make the namer consider reserved words like function as "used"

the logic is really straightforward, the only complicated thing here is how this logic is plugged in to the templates (which, as you surely know by now, don't allow much to do )
```

sofisl:
Fixes googleapis/google-cloud-node-core#583


fhinkel (approved):

summer-ji-eng (approved):

alexander-fenster (approved):

bcoe:
What if rather than `innerFunctionName`, we went with `callFunctoinName`, `inner` seemed slightly weird to me because it's actually the outer function from the perspective of an embeded sample.

bcoe (changes requested):
See recommendations.

bcoe (approved):

PR #1041, opened 2021-11-02
fix: replace empty objects with curly brackets instead of strings

### LINKED ISSUE #593: snippet-gen: object variable appears to default to "".
If a parameter is an object, it would be less confusing if we defaulted to `{}`, in the case below can the parameter be a string or object? or did we just miss applying this logic:

Screen Shot 2021-09-22 at 6 54 31 AM

Refs: https://github.com/googleapis/nodejs-logging/pull/1160

### LINKED ISSUE #980: snippet-gen: object variable appears to default to "".
If a parameter is an object, it would be less confusing if we defaulted to `{}`, in the case below can the parameter be a string or object? or did we just miss applying this logic:

Screen Shot 2021-09-22 at 6 54 31 AM

Refs: https://github.com/googleapis/nodejs-logging/pull/1160

sofisl:
Fixes googleapis/google-cloud-node-core#593

### DIFF
--- a/baselines/asset/samples/generated/v1/asset_service.batch_get_assets_history.js.baseline
+++ b/baselines/asset/samples/generated/v1/asset_service.batch_get_assets_history.js.baseline
@@ -41,7 +41,7 @@ function main(parent) {
   /**
    *  Optional. The content type.
    */
-  // const contentType = ''
+  // const contentType = {}
   /**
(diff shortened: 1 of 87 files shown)


bcoe (commented):
👍 I'm really happy with this fix.

bcoe (approved):

summer-ji-eng (approved):

PR #1066, opened 2021-12-04
fix: add null request for bi_di stream call

### LINKED ISSUE #425: bi-directional streaming calls with extra options
Based on the discussion [here](https://github.com/googleapis/nodejs-firestore/pull/1019#discussion_r409169313):

```diff
- return this.innerApiCalls.listen(options);
+ return this.innerApiCalls.listen({}, options);
```

@schmidt-sebastian wrote:
> I do think the generated code is busted here. We use options to set an additional header, but the way this gets passed down here is as the request (if I remember this correctly). If we don't modify the code, the header is dropped. This only affects bidirectional streaming APIs with additional metadata, so probably just us. I do think we should fix this upstream at some point.

We need to figure out what's wrong there, and fix it in the generator.

### LINKED ISSUE #657: bi-directional streaming calls with extra options
Based on the discussion [here](https://github.com/googleapis/nodejs-firestore/pull/1019#discussion_r409169313):

```diff
- return this.innerApiCalls.listen(options);
+ return this.innerApiCalls.listen({}, options);
```

@schmidt-sebastian wrote:
> I do think the generated code is busted here. We use options to set an additional header, but the way this gets passed down here is as the request (if I remember this correctly). If we don't modify the code, the header is dropped. This only affects bidirectional streaming APIs with additional metadata, so probably just us. I do think we should fix this upstream at some point.

We need to figure out what's wrong there, and fix it in the generator.

summer-ji-eng:
Fixes googleapis/google-cloud-node-core#657.

`appendRows` bi-di stream call in `nodejs-bigquery-storage` fail to pass in the options.
```
        let options = {}
        options.otherArgs = {};
        options.otherArgs.headers = {};
        options.otherArgs.headers['x-goog-request-params'] = writeStream
        const stream = await storageClient.appendRows(options);
```

Because bi-direction stream call treat `options` as `request` when [`createAPICall` ]https://github.com/googleapis/gax-nodejs/blob/be8175624624648a32d535c3f6495814c163eb18/src/createApiCall.ts#L73)

Fix: Similar to [client stream](https://github.com/googleapis/gapic-generator-typescript/blob/main/templates/typescript_gapic/src/%24version/%24service_client.ts.njk#L660), we could pass `null` as request before `options`.
```
return this.innerApiCalls.{{ method.name.toCamelCase() }}(null, options);
```

### DIFF
--- a/baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
+++ b/baselines/disable-packing-test/src/v1beta1/echo_client.ts.baseline
@@ -593,7 +593,7 @@ export class EchoClient {
       options?: CallOptions):
     gax.CancellableStream {
     this.initialize();
-    return this.innerApiCalls.chat(options);
+    return this.innerApiCalls.chat(null, options);
   }
 
 /**
--- a/baselines/disable-packing-test/src/v1beta1/messaging_client.ts.baseline
+++ b/baselines/disable-packing-test/src/v1beta1/messaging_client.ts.baseline
@@ -1034,7 +1034,7 @@ export class MessagingClient {
       options?: CallOptions):
     gax.CancellableStream {
     this.initialize();
-    return this.innerApiCalls.connect(options);
+    return this.innerApiCalls.connect(null, options);
   }
 
 /**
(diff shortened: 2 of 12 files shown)


alexander-fenster (approved):
Looks good to me! I suggest to merge it but not release until the new year though, since this fix has some potential of breaking the existing code.

summer-ji-eng:
> Looks good to me! I suggest to merge it but not release until the new year though, since this fix has some potential of breaking the existing code.

Thanks for suggestion. Will mark the new release with `DO NOT MERGE` tag.

PR #1076, opened 2021-12-16
fix: ignore generating non-diregapic pagination with map response

summer-ji-eng:
Ignore the non diregapic pagination methtod where the response contain maps.
```
message PagedExpandLegacyMappedResponse {
  // The words that were expanded, indexed by their initial character.
  // (-- aip.dev/not-precedent: This is a legacy, non-standard pattern that violates
  //     aip.dev/158. Ordinarily, this should be a `repeated` field, as in PagedExpandResponse. --)
  map<string, PagedExpandResponseList> alphabetized = 1;

  // The next page token.
  string next_page_token = 2;
}
```
See `PagedExpandLegacyMapped` in [showcase proto](https://github.com/googleapis/gapic-showcase/blob/master/schema/google/showcase/v1beta1/echo.proto#L101-L106)

### DIFF
--- a/baselines/compute/test/gapic_addresses_v1.ts.baseline
+++ b/baselines/compute/test/gapic_addresses_v1.ts.baseline
@@ -343,7 +343,6 @@ describe('v1.AddressesClient', () => {
                 .getCall(0).calledWith(request, expectedOptions, undefined));
         });
     });
-
     describe('aggregatedList', () => {
 
         it('uses async iteration with aggregatedList without error', async () => {
@@ -404,7 +403,6 @@ describe('v1.AddressesClient', () => {
             );
         });
     });
-
     describe('list', () => {
         it('invokes list without error', async () => {
             const client = new addressesModule.v1.AddressesClient({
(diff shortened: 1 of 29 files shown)


baselines/compute/test/gapic_addresses_v1.ts.baseline
```diff
@@ -343,7 +343,6 @@ describe('v1.AddressesClient', () => {
                 .getCall(0).calledWith(request, expectedOptions, undefined));
         });
     });
-
```
alexander-fenster, line 346: Maybe it's better to put this empty line back if not too hard?

typescript/src/schema/proto.ts
```diff
@@ -316,6 +317,24 @@ function pagingResponseType(
   return '.google.protobuf.FieldDescriptorProto.Type.' + type;
 }
 
+// Ignore non-diregapic pagation method where its response type contains a map.
+function ignorePagingMethod(
```
alexander-fenster, line 321: Let's rename it to `ignoreMapPagingMethod` here and everywhere, otherwise the name is confusing.

alexander-fenster (approved):

baselines/compute/test/gapic_addresses_v1.ts.baseline
summer-ji-eng: Agree.

typescript/src/schema/proto.ts
summer-ji-eng: agree

PR #1091, opened 2022-01-21
fix: change dynamic routing header field to camelCase

alicejli:
Fix for Issue googleapis/google-cloud-node-core#579 

1) Fixed issue where the routing annotation request field was not being converted into camelCase (thus causing issues as Javascript converts all the fields to camelCase)
2) Added more unit tests to cover above use case + nested field case

### DIFF
--- a/templates/typescript_gapic/_util.njk
+++ b/templates/typescript_gapic/_util.njk
@@ -261,9 +261,9 @@ request.{{ oneComment.paramName.toCamelCase() }}
 {%- if method.dynamicRoutingRequestParams.length > 0 -%}
     let routingParameter = {};
 {%- for paramArray in method.dynamicRoutingRequestParams %}
-  {%- for sameParamArray in paramArray %}
-      if(RegExp('{{ sameParamArray.messageRegex | safe }}').test(request.{{ sameParamArray.fieldRetrieve }}!)){
-        Object.assign(routingParameter, { {{ sameParamArray.fieldSend }}: request.{{ sameParamArray.fieldRetrieve }}!.match(RegExp('{{ sameParamArray.namedSegment | safe }}'))![0]})
+  {%- for param in paramArray %}
+      if(RegExp('{{ param.messageRegex | safe }}').test(request.{{ param.fieldRetrieve }}!)){
+        Object.assign(routingParameter, { {{ param.fieldSend }}: request.{{ param.fieldRetrieve }}!.match(RegExp('{{ param.namedSegment | safe }}'))![0]})
       } else
   {%- endfor %}
       {}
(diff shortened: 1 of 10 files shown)


templates/typescript_gapic/_util.njk
```diff
@@ -262,7 +262,7 @@ request.{{ oneComment.paramName.toCamelCase() }}
     let routingParameter = {};
 {%- for paramArray in method.dynamicRoutingRequestParams %}
   {%- for sameParamArray in paramArray %}
```
summer-ji-eng, line 264: nit: could we rename `sameParamArray` as `param` for simplicity. go/tott/431 is good reference.

typescript/test/unit/proto.ts
```diff
@@ -251,6 +252,25 @@ describe('src/schema/proto.ts', () => {
...
+      assert.deepStrictEqual(convertFieldToCamelCase(''), '');
+      assert.deepStrictEqual(convertFieldToCamelCase('parent_id'), 'parentId');
+      assert.deepStrictEqual(
+        convertFieldToCamelCase('app_profile_id'),
+        'appProfileId'
+      );
+      assert.deepStrictEqual(
+        convertFieldToCamelCase('name.parent_id.another_parent_id'),
+        'name.parentId.anotherParentId'
+      );
```
summer-ji-eng, line 270: Can we add a unit test on a field name with non-alphanumeric letter like `name?_(1)`. I expect the result is same `name?_(1)`.

summer-ji-eng (approved):
Overall LGTM, with details comment for readability and simplicity. Feel free to merge the PR after address the comments.

Thank you for your promptly  fix. 👍

templates/typescript_gapic/_util.njk
alicejli: Makes sense - thanks for the reference!

typescript/test/unit/proto.ts
alicejli: I've added it, but based on how the `toCamelCase` function is defined, `name?_(1)` actually would get converted to `name_1`, because it splits on non-alphanumeric characters. This should be okay in our case.

in `util.ts`:

```
String.prototype.words = function (this: string): string[] {
  // split on spaces, non-alphanumeric, or capital letters
  return this.split(/(?=[A-Z])|[\s\W_]+/)
    .filter(w => w.length > 0)
    .map(w => w.toLowerCase());
};

String.prototype.toCamelCase = function (this: string): string {
  const words = this.words();
  if (words.length === 0) {
    return this;
  }
  const result = [words[0]];
  result.push(
    ...words.slice(1).map(w => {
      if (isDigit(w)) {
        return '_' + w;
      }
      return w.capitalize();
    })
  );
  return result.join('');
};
```

summer-ji-eng: `name?_(1)` -> `name_1` is not correct. But let's move the test case for now. I hope that no client use such wired name. If it comes later, I will take care of it . Thank you.

alicejli: Got it, sounds good. Removed.

PR #1098, opened 2022-01-31
fix: add check for undefined routing header parameters

alicejli:
This adds a check for dynamic routing annotations based on fields that may be undefined.

TODO: Add integration test to generator to test dynamic routing header with gapic-showcase.


summer-ji-eng:
Let's track the integration test in the issues. LGTM

summer-ji-eng (approved):

PR #1101, opened 2022-02-02
fix: close client with stream methodshould return stream object

summer-ji-eng:
Problem: client library always return a rejected promise if the client has been closed. However, when stream make an apicall, it expect a stream object not promise object. This cause `TypeError` of `.on` is not function.

Fix: return a stream with error event if ongoing stream make on closed client.


summer-ji-eng:
Investigate if we need a unit test on this case.

summer-ji-eng:
Bidi stream and client stream return different error from grpc like:
```
{"code":14,"details":"No connection established","metadata":{}}
```

baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
```diff
@@ -217,6 +218,13 @@ export class BigQueryStorageClient {
       const callPromise = this.bigQueryStorageStub.then(
         stub => (...args: Array<{}>) => {
           if (this._terminated) {
+            if (methodName in this.descriptors.stream) {
+              const stream = new PassThrough();
+              setImmediate(() => {
+                stream.emit('error', new GoogleError('The client has already been closed.'));
+              });
+              return stream;
+            }
```
igorbernstein2, line 227: doesn't this need to also happen in the rejection arm? ie if the `bigQueryStorageStub` is rejected (ie. no creds), the the promise will still get bubbled up to streaming.js

igorbernstein2 (changes requested):

baselines/bigquery-storage/src/v1beta1/big_query_storage_client.ts.baseline
summer-ji-eng: If the `bigQueryStorageStub` is rejected, it throw an error(see line 225 below). This wouldn't get bubble up to streaming.js.

In other words, only when`bigQueryStorageStub` resolved, and client was terminated, the streaming call would receive `Passthrough` stream with error event.

igorbernstein2: ok

templates/typescript_gapic/test/gapic_$service_$version.ts.njk
```diff
@@ -517,6 +517,32 @@ describe('{{ api.naming.version }}.{{ service.name }}Client', () => {
             assert((client.innerApiCalls.{{ method.name.toCamelCase() }} as SinonStub)
                 .getCall(0).calledWith(request, expectedOptions));
         });
+
+        it('invokes {{ method.name.toCamelCase() }} with closed client', async () => {
```
alexander-fenster, line 521: Thanks for covering this case in the generated unit test!

templates/typescript_gapic/test/gapic_$service_$version.ts.njk
```diff
@@ -517,6 +517,32 @@ describe('{{ api.naming.version }}.{{ service.name }}Client', () => {
...
+            const client = new {{ service.name.toLowerCase() }}Module.{{ api.naming.version }}.{{ service.name }}Client({{- util.initClientOptions(api.rest) -}});
+            {%- if method.options and method.options.deprecated %}
+            const stub = sinon.stub(client, 'warn');
+            {%- endif %}
+            client.{{ id.get("initialize") }}();
+            const request = generateSampleMessage(new protos{{ method.inputInterface }}());
+            const expectedError = new Error('The client has already been closed.');
+            setTimeout(() => {
+                client.close();
+            }, 1000);
```
alexander-fenster, line 531: I'm afraid 1 second is too long in this case since the test will likely complete immediately. Does it actually work with 1 second?  I would probably do `setImmediately`, or maybe even just `client.close()` synchronously.

alexander-fenster (approved):
Please don't release it to all APIs, let's just generate bigtable with this change and see if it works for them, then we can release it across the board.

templates/typescript_gapic/test/gapic_$service_$version.ts.njk
summer-ji-eng: Update to `client.close()` synchronously.

PR #1106, opened 2022-02-10
feat: add note to autogenerated samples for handwritten tags

sofisl:


templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -512,8 +512,10 @@ export class {{ service.name }}Client {
 {%- for method in service.simpleMethods %}
 /**
 {{- util.printComments(method, service) }}
+ {%- if not api.handwrittenLayer %}
```
bcoe, line 515: how is the `api.handwrittenLayer` variable set?

summer-ji-eng (approved):

templates/typescript_gapic/src/$version/$service_client.ts.njk
summer-ji-eng: I remember the variable set through bazel rules ([example](https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/BUILD.bazel#L220))

Please correct me @sofisl if I am wrong.

sofisl: Yep, exactly! We haven't added it yet, but it would be set there, where Summer says ^

feywind:
Cool, sounds good :)

bcoe (approved):

kolea2 (approved):

PR #1119, opened 2022-03-04
fix: should ignore google/rpc/context/*.proto

summer-ji-eng:
Address googleapis/google-cloud-node-core#575


typescript/src/templater.ts
```diff
@@ -60,8 +60,6 @@ function createSnippetIndexMetadata(
   api: API,
   basePath: string
 ): protos.google.cloud.tools.snippetgen.snippetindex.v1.IIndex {
-  const apis: protos.google.cloud.tools.snippetgen.snippetindex.v1.IApi[] = [];
```
summer-ji-eng, line 63: fix linter `the api variable never been used`

alexander-fenster (approved):

PR #1123, opened 2022-03-08
fix: update comment for diregapic LRO method with map response

summer-ji-eng:


templates/typescript_gapic/_util.njk
```diff
@@ -74,8 +74,12 @@ limitations under the License.
 
 
 {%- macro printPageAsyncHeader(method) %}
+{%- if method.pagingMapResponseType -%}
+{{ printCommentsForMethod(method) }}
```
vchudnov-g, line 78: Is the idea that for the methods where we say they're equivalent to some other method, we don't repeat the method comments (ie don't call  `printCommentsForMethod`), but in method where we don't refer to a another methods we do insert the original comment? (Just making sure I understand why this call is in the `if` and not in the `else`).

summer-ji-eng: Exactly.

`printCommentsForMethod(method)` => is printing the comments above the RPC definition in protos(.i.e [aggregatedList](https://github.com/googleapis/googleapis/blob/master/google/cloud/compute/v1small/compute_small.proto#L671)). It only print for diregapic LRO with map response. Otherwise, we say "equivalent to some other method".

vchudnov-g (approved):

alexander-fenster (approved):

PR #1124, opened 2022-03-10
fix: rename path template params which is reserved word in strict mode

summer-ji-eng:
1. `nodejs-security-center` uses `export` keyword, preventing [compile](https://github.com/googleapis/nodejs-security-center/pull/467)
2. `nodejs-artifact-registry` uses `package` keyword preventing [compile](https://github.com/googleapis/nodejs-artifact-registry/pull/121).

TS complains the key words error(see below) when the generator uild a list of resources referenced by this service. We rename the these key words in the [strict mode reserved words](https://github.com/microsoft/TypeScript/issues/2536):

| Strict Mode Reserved Words |
| -- |
| as |
implements
interface
let
package
private
protected
public
static
yield

```
Identifier expected. 'package' is a reserved word in strict mode. Class definitions are automatically in strict mode.
```


typescript/src/schema/proto.ts
```diff
@@ -33,6 +33,19 @@ const COMMON_PROTO_LIST = [
   'google.type',
 ];
 
+const STRICT_MODE_RESERVED_WORDS = new Set<String>([
```
alexander-fenster, line 36: Sorry, this cannot be here since it is specific for TypeScript and the generator code must be language independent by design (otherwise, all possible issues will arise).  The only possible place for this logic is the `namer.js` in the template. Good news is that it should be pretty straightforward to move it there.

summer-ji-eng: Thanks for the advice. That's smart design. New lesson learned.

templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -913,33 +913,33 @@ export class {{ service.name }}Client {
...
-      {{ param }}: {{ param.toCamelCase() }},
+      {{ id.get(param) }}: {{ id.get(param.toCamelCase()) }},
 {%- endfor %}
     });
   }
 {%- for param in template.params %}
 
   /**
-   * Parse the {{ param }} from {{ template.name.toPascalCase() }} resource.
+   * Parse the {{ id.get(param) }} from {{ template.name.toPascalCase() }} resource.
```
alexander-fenster, line 935: I think it can stay `param` in the comment, we only need to rename it in places where it's a variable name.

alexander-fenster: It's hard to figure out what's going on :)  We have a special baseline test to check the renames ([naming](https://github.com/googleapis/gapic-generator-typescript/blob/main/typescript/test/unit/baselines.ts#L136-L142)), can you add a conflicting template parameter name to see how it works in the baseline?

Thank you!

templates/typescript_gapic/namer.js
```diff
@@ -20,6 +20,19 @@
  * the template engine.
  */
 
+const STRICT_MODE_RESERVED_WORDS = new Set([
```
bcoe, line 23: Did we recently turn on strict mode, or did we just happen to have not bumped into this issue before?

summer-ji-eng: As I know, we always use `strict mode`. I think it just bumped this issue this time.

templates/typescript_gapic/src/$version/$service_client.ts.njk
summer-ji-eng: Keep `param` in the comment.

Added `Resource` with reserved word in `naming.proto`

baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -1109,6 +1186,81 @@ export class NamingClient {
...
+   * Return a fully-qualified projectExport resource name string.
+   *
+   * @param {string} project
+   * @param {string} exportParam
+   * @returns {string} Resource name string.
+   */
+  projectExportPath(project:string,exportParam:string) {
+    return this.pathTemplates.projectExportPathTemplate.render({
+      project: project,
+      exportParam: exportParam,
```
alexander-fenster, line 1203: I believe it should say `export` here, otherwise it won't get rendered properly.  You can do `"export"`, `gts` will remove the quotes if they are not needed.

summer-ji-eng: Cool. After couple trys with gax and even created a useless [PR](https://github.com/googleapis/gax-nodejs/pull/1202) in gax😢, finally, I got the idea. I only need to update the identifier in the method parameter. Update the PR and please take another look. Thanks for bearing with me. 🤝

alexander-fenster (approved):

PR #1130, opened 2022-04-23
fix: dynamic route fix

summer-ji-eng:
Now generator only works for multiple variables in HTTP URI in format like `{param1.param2.param3=abc/*/d/*/ef/}`. It doesn't support format in `/v1/projects/{project_id}/locations/{location}/jobs`.

[dataflow proto](https://github.com/googleapis/googleapis/blob/master/google/dataflow/v1beta3/jobs.proto#L105-L112) has recent change. Because generator only find the first HTTP request param `project_id`([client library code](https://github.com/googleapis/nodejs-dataflow/blob/main/src/v1beta3/jobs_v1_beta3_client.ts#L894-L897
)), it cause `loctation` error. The reason is that the routing header misses `location` to sent `x-goog-request-param` header in grpc matedata. 

To sent the correct the http request parameter, the client library surface code should has change like this:
```
options.otherArgs.headers['x-goog-request-params'] =
      gax.routingHeader.fromParams({
        project_id: request.projectId || '',
        location: request.location || '',
      });
```

### DIFF
--- a/typescript/src/schema/proto.ts
+++ b/typescript/src/schema/proto.ts
@@ -568,13 +568,6 @@ function augmentMethod(
     method.headerRequestParams = getHeaderRequestParams(
       method.options?.['.google.api.http']
     );
-    if (method.name === 'ListJobs') {
-      console.warn('--- method.name:: ', method.name)
-      console.warn('--- method.option?.[.google/api/http]', method.options?.['.google.api.http'])
-      console.warn('--method.headerRequestParams:: ', method.headerRequestParams)
-    }
-
-    // console.warn('--- method.headerRequestParams:: ', method.headerRequestParams)
   }
   // If dynamic routing annotation exists and is non-empty, then send dynamic routing headers.
   else if (
--- a/typescript/src/schema/proto.ts
+++ b/typescript/src/schema/proto.ts
@@ -612,7 +612,6 @@ export function getHeaderRequestParams(
   params = params.concat(
     ...httpRule.additionalBindings.map(binding => getSingleHeaderParam(binding))
   );
-
   // de-dup result array
   const used = new Set();
   const result: string[][] = [];
--- a/typescript/src/schema/proto.ts
+++ b/typescript/src/schema/proto.ts
@@ -593,7 +593,9 @@ export function getSingleHeaderParam(
     const result: string[][] = [];
     const matches = message.matchAll(/{(.*?)[=}]/g);
     for (const match of matches) {
-      result.push(match[1] ? match[1].split('.') : []);
+      if (match[1]) {
+        result.push(match[1].split('.'));
+      }
     }
     return result;
   }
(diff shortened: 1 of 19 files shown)


typescript/src/schema/proto.ts
```diff
@@ -586,12 +586,16 @@ function augmentMethod(
...
   rule: protos.google.api.IHttpRule
-): string[] {
+): string[][] {
   const message =
     rule.post || rule.delete || rule.get || rule.put || rule.patch;
   if (message) {
-    const res = message.match(/{(.*?)[=}]/);
-    return res?.[1] ? res[1].split('.') : [];
+    const result: string[][] = [];
+    const matches = message.matchAll(/{(.*?)[=}]/g);
```
alexander-fenster, line 594: note: `matchAll` appears in Node.js v12 so you need to update `engines` filed in `package.json` and drop Node 10 support for the generator to be able to safely use it. That's OK, since it's just a generator dependency, the client libraries are not affected by this.

typescript/src/schema/proto.ts
```diff
@@ -586,12 +586,16 @@ function augmentMethod(
...
+    for (const match of matches) {
+      result.push(match[1] ? match[1].split('.') : []);
```
alexander-fenster, line 596: Why do we need to push an empty array here? Just push `if (match[1]) { result.push(...); }`

typescript/src/schema/proto.ts
```diff
@@ -586,12 +586,16 @@ function augmentMethod(
 
 export function getSingleHeaderParam(
   rule: protos.google.api.IHttpRule
-): string[] {
+): string[][] {
```
alexander-fenster, line 589: Is there any reason why it must be a 2d array and not just a string array of all matches?

typescript/src/schema/proto.ts
summer-ji-eng: Good point. Will update on this.

typescript/src/schema/proto.ts
summer-ji-eng: Thanks for catching this up. I certainly upgrade to v12 in generator.

typescript/src/schema/proto.ts
summer-ji-eng: This method is used to support multiple variable in format like `'{param1.param2=abc/*/foo/*/}`. This method separate by "." and return array like `[param1, param2]`.

After 5 minutes thought, I think we could flat the 2 d array. Let me try out.

summer-ji-eng:
Test should pass once #1132 merged.

typescript/src/schema/proto.ts
alexander-fenster: I see now, thank you!

alexander-fenster (approved):

PR #1149, opened 2022-06-17
fix: camel case for multiple capital letters

### LINKED ISSUE #570: bug: searching for method names incorrectly when there's an acronym
1. gRPC reads proto definitions and generates an object that, for each rpc in the proto definition, defines a method that would call that RPC for 
2. so if your proto has rpc SomeFunc(Request) returns (Response), and you pass this proto to gRPC, it will return an object (that we call a stub) that would have stub["someFunc"], note the camelCase
3. When stubs are generated for methods that have an acronym in them, they get generated like such: `createOsPolicyAssignment`, but the generated code looks for the method name `createOSPolicyAssignment`

We need to make sure we're looking for the right method name.

See: https://github.com/googleapis/nodejs-os-config/issues/173

### LINKED ISSUE #1148: bug: searching for method names incorrectly when there's an acronym
1. gRPC reads proto definitions and generates an object that, for each rpc in the proto definition, defines a method that would call that RPC for 
2. so if your proto has rpc SomeFunc(Request) returns (Response), and you pass this proto to gRPC, it will return an object (that we call a stub) that would have stub["someFunc"], note the camelCase
3. When stubs are generated for methods that have an acronym in them, they get generated like such: `createOsPolicyAssignment`, but the generated code looks for the method name `createOSPolicyAssignment`

We need to make sure we're looking for the right method name.

See: https://github.com/googleapis/nodejs-os-config/issues/173

alexander-fenster:
Fixes googleapis/google-cloud-node-core#570, https://github.com/googleapis/nodejs-os-config/issues/173.

`protobuf.js` and `@grpc/grpc-js` use somewhat different logic when camel-casing strings like `CreateOSPolicy`, where multiple capital letters form a word. In our current code, it will be camel-cased as `createOSPolicy`, while these libraries expect `createOsPolicy`.

Implementing a special mode to use internally for this specific case. External interfaces of the generated libraries should not be affected in any way.


baselines/naming/src/v1beta1/naming_client.ts.baseline
```diff
@@ -240,7 +240,7 @@ export class NamingClient {
     // Iterate over each of the methods that the service provides
     // and create an API call method for each.
     const namingStubMethods =
-        ['paginatedMethod', 'paginatedMethodStream', 'paginatedMethodAsync', 'longRunning', 'checkLongRunningProgress', 'initialize', 'servicePath', 'apiEndpoint', 'port', 'scopes', 'getProjectId', 'getReservedWord'];
+        ['paginatedMethod', 'paginatedMethodStream', 'paginatedMethodAsync', 'longRunning', 'checkLongRunningProgress', 'initialize', 'servicePath', 'apiEndpoint', 'port', 'scopes', 'getProjectId', 'getReservedWord', 'createAbcdeSomething'];
```
alexander-fenster, line 243: Here is the main change: if you scroll to the right, the last element of this list is `createAbcdeSomething`.  It would've been `createABCDESomething` without the fix.

sofisl (approved):

PR #1160, opened 2022-07-19
fix: proper processing of routing headers

### LINKED ISSUE #569: cloud run API: linting error in the generated code
image

```
src/v2/services_client.ts#L1035
Optional chain expressions can return undefined by design - using a non-null assertion is unsafe and wrong
```

### LINKED ISSUE #1154: cloud run API: linting error in the generated code
image

```
src/v2/services_client.ts#L1035
Optional chain expressions can return undefined by design - using a non-null assertion is unsafe and wrong
```

alexander-fenster:
Fixes googleapis/google-cloud-node-core#569.

A huge refactor of the routing header processing logic to fix the Cloud Run use case. Tested locally, the generated Cloud Run client works (successfully providing the `location` routing header).

### DIFF
--- a/baselines/routingtest/src/v1/test_service_client.ts.baseline
+++ b/baselines/routingtest/src/v1/test_service_client.ts.baseline
@@ -342,14 +342,22 @@ export class TestServiceClient {
     options = options || {};
     options.otherArgs = options.otherArgs || {};
     options.otherArgs.headers = options.otherArgs.headers || {};let routingParameter = {};
-      if ((typeof request.name !== "undefined") && RegExp('(?<database>projects)/[^/]+/databases/[^/]+/documents').test(request.name!)){
-        Object.assign(routingParameter, { database: request.name!.match(RegExp('(?<database>projects/[^/]+/databases/[^/]+)'))![0]})}
+    {
+      const fieldValue = request.name;
+      if (fieldValue !== undefined && fieldValue !== null) {
+        const match = fieldValue.toString().match(RegExp('(?<database>projects/[^/]+/databases/[^/]+)/documents'));
+        if (match) {
+          const parameterValue = match.groups?.['database'] ?? fieldValue;
+          Object.assign(routingParameter, { database: parameterValue });
+        }
+      }
+    }
 
-        options.otherArgs.headers[
-          'x-goog-request-params'
-          ] = gax.routingHeader.fromParams(
-            routingParameter
-          );
+    options.otherArgs.headers[
+      'x-goog-request-params'
+    ] = gax.routingHeader.fromParams(
+      routingParameter
+    );
     this.initialize();
     return this.innerApiCalls.fastFibonacci(request, options, callback);
   }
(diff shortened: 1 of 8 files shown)


typescript/src/util.ts
```diff
@@ -221,144 +221,35 @@ export function getResourceNameByPattern(pattern: string): string {
...
-): boolean {
-  let countOfNamedSegments = 0;
-  // Checks that resources are named
-  const regexNamed = new RegExp(/^.+=.+/);
-  const regexWildcard = new RegExp(/^{[a-zA-Z\d-]+}/);
-  pattern.forEach(element => {
-    if (regexNamed.test(element) || regexWildcard.test(element)) {
-      countOfNamedSegments++;
-    }
+export function processPathTemplate(
```
sofisl, line 224: Would you mind adding a comment for this function? I'm not quite sure what it does, or how it replaces all the deleted functions in TS.

typescript/src/schema/proto.ts
```diff
@@ -703,19 +701,20 @@ export function getSingleRoutingHeaderParam(
...
-      messageRegex: '[^/]+',
-      namedSegment: '[^/]+',
+      messageRegex: `(?<${rule.field}>.*)`,
     };
   }
   // If the annotation is malformed, then return empty interface
-  else if (getNamedSegment(rule.pathTemplate).length < 1) {
-    return dynamicRoutingRule;
-  } else {
+  else {
```
sofisl, line 708: I assume that if the annotation is malformed, those values don't get appended to the header. What happens to the request in that case?

baselines/routingtest/src/v1/test_service_client.ts.baseline
```diff
@@ -422,17 +430,33 @@ export class TestServiceClient {
...
     options.otherArgs = options.otherArgs || {};
     options.otherArgs.headers = options.otherArgs.headers || {};let routingParameter = {};
-      if ((typeof request.name !== "undefined") && RegExp('[^/]+').test(request.name!)){
-        Object.assign(routingParameter, { name: request.name!.match(RegExp('[^/]+'))![0]})}
+    {
+      const fieldValue = request.name;
+      if (fieldValue !== undefined && fieldValue !== null) {
+        const match = fieldValue.toString().match(RegExp('(?<name>.*)'));
+        if (match) {
+          const parameterValue = match.groups?.['name'] ?? fieldValue;
```
sofisl, line 438: are these changes safe to make to other APIs? (i.e., can we/should we send those values in the header for APIs other than cloud run?)

typescript/src/schema/proto.ts
alexander-fenster: At this moment, only two APIs (BigTable and CloudRun) use this functionality, and I checked that everything works :)

typescript/src/util.ts
alexander-fenster: Sure!

typescript/src/schema/proto.ts
alexander-fenster: Most likely it will fail on the backend. The client library is not supposed to validate the request, it's OK to let it through and let the backend respond.

sofisl (approved):

PR #1211, opened 2022-08-17
fix(samples): add disclaimer

sofisl:
Fixes b/240118972 internally

Labeling it as a feat so that it releases a new version of the generator, but will send the individual PRs out as docs


alexander-fenster (approved):

alexander-fenster:
`fix` should be enough to trigger a release here.

PR #1251, opened 2022-09-08
feat: option to request REST numeric enum values

alexander-fenster:
Passing an option to turn on numeric enums for REST transport down the stack.


templates/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -157,6 +157,11 @@ export class {{ service.name }}Client {
     const fallback = opts?.fallback ?? (typeof window !== 'undefined' && typeof window?.fetch === 'function');
     opts = Object.assign({servicePath, port, clientConfig, fallback}, opts);
     {%- endif %}
+    {%- if api.restNumericEnums %}
```
vchudnov-g, line 160: Is this inside something like `if rest` so it only gets applied when we generate REST transport?

vchudnov-g (commented):
Looks good. Just one question, which may be due to my not being familiar with the TS generator.

vchudnov-g (approved):
LGTM, though I don't claim expertise on the TS generator.

templates/typescript_gapic/src/$version/$service_client.ts.njk
alexander-fenster: As discussed - we generate almost the same code for REST and gRPC client (the only difference being the default transport), so we want to pass this option down to gax even if the client is generated as a gRPC one, because the customer might choose to use it as a REST client by passing `{fallback: 'rest'}` to the constructor.

PR #1307, opened 2023-02-01
feat!: convert to ESM, use Aspect build rules

alexander-fenster:
This PR consists of four commits that are logically unrelated but should go together. It might be easier to review them one by one.

1. Update baselines (copyright year and minor changes of dependency protos). This one is straightforward.

2. Change the Bazel stack to use the new Aspect rules for JavaScript and TypeScript. Unfortunately, for TypeScript, I'm currently using my fork of the rules that fixes some issues; I'll try to contribute it upstream but I'm willing to use the fork while this is in progress to unblock this repository.

3. Make the generator an ESM module. It simplifies working with new dependencies, and make TypeScript emit more predictable code (compared with the one with commonjs requires).

4. Remove fragile parts that keep breaking. We are getting rid of `gts` and `eslint` (for now, until they properly support the new ESM import syntax - I filed googleapis/google-cloud-node-core#560) and also getting rid of complicated integration tests: they were broken anyway but were reporting pass, which is the worst thing a test can do. I'll need to rethink how to make good integration tests. For now, we still have baseline test and we check that the generated libraries compile properly.

R= @noahdietz for Bazel related fixes ([this](https://github.com/googleapis/gapic-generator-typescript/commit/08fdeb29082d44cc1d2d8247fcb4a639ccf7605c) commit specifically)

R= @sofisl and @bcoe for ESM related fixes ([this](https://github.com/googleapis/gapic-generator-typescript/commit/308702115f1b9f5794d871b2b51ac45de1e60732) commit)


alexander-fenster:
@ddelgrosso1 FYI - Ben suggested I add you for the ESM conversion commit (just [this one](https://github.com/googleapis/gapic-generator-typescript/pull/1307/commits/308702115f1b9f5794d871b2b51ac45de1e60732)). This is our GAPIC generator and it's going to be an ESM app from now on. The fact that it's built by Bazel greatly complicates things.

typescript/src/util.ts
```diff
@@ -258,3 +227,43 @@ export function processPathTemplate(
 
   return {fieldSend, messageRegex};
 }
+
+// Quick'n'dirty implementation of mkdirp to avoid pulling a dependency
+export async function mkdirp(directory: string) {
```
bcoe, line 232: Node 14 should have mkdirp built in, just set `recursive = true` for any `mkdir` operations.

bcoe (changes requested):
Left a couple nits, I think there are some built in Node.js methods you could be using like `mkdir` with the recursive flag.

package.json
```diff
@@ -27,7 +28,9 @@
...
     "compile": "bazel build //...",
-    "compile-protos": "pbjs -t static-module -o protos/index.js -p node_modules/google-gax/build/protos -p protos google/api/annotations.proto google/api/field_behavior.proto google/api/resource.proto google/api/routing.proto google/longrunning/operations.proto google/protobuf/compiler/plugin.proto service_config.proto snippet_index.proto && pbts protos/index.js -o protos/index.d.ts",
+    "compile-protos-json": "pbjs -t json -o protos/protos.json -p node_modules/google-gax/build/protos -p protos google/api/annotations.proto google/api/field_behavior.proto google/api/resource.proto google/api/routing.proto google/longrunning/operations.proto google/protobuf/compiler/plugin.proto service_config.proto snippet_index.proto",
+    "compile-protos-js": "pbjs -t static-module -o protos/index.js -p node_modules/google-gax/build/protos -p protos google/api/annotations.proto google/api/field_behavior.proto google/api/resource.proto google/api/routing.proto google/longrunning/operations.proto google/protobuf/compiler/plugin.proto service_config.proto snippet_index.proto",
```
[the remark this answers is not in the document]
alexander-fenster: As in libraries, we need JSON for the code to run (we read it in the runtime), and we need JS to produce types. We are only supposed to run these commands when the protos are changed significantly.

typescript/src/util.ts
alexander-fenster: Oh nice! I target Node 18 so I'll make this change now, thank you!

typescript/test/test-application/test-js.ts
```diff
@@ -66,15 +66,15 @@ describe('Test application for JavaScript users', () => {
   it('npm install showcase', async function () {
     this.timeout(240000);
     // copy protos to generated client library and copy test application to local.
-    fs.copySync(protos, path.join(showcaseLib, 'protos'));
-    fs.copySync(jsTestApplication, localJsApplication);
+    await copy(protos, path.join(showcaseLib, 'protos'));
```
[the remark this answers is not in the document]
alexander-fenster: Awesome!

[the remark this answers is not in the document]
alexander-fenster: Changed to recursive operation.

typescript/src/schema/proto.ts
```diff
@@ -393,22 +392,26 @@ function getMethodConfig(
...
   let exactMatch: protos.grpc.service_config.IMethodConfig | undefined;
   let serviceMatch: protos.grpc.service_config.IMethodConfig | undefined;
-  for (const config of grpcServiceConfig.methodConfig) {
-    if (!config.name) {
-      continue;
-    }
-    for (const name of config.name) {
-      if (name.service === serviceName && !name.method) {
-        serviceMatch = config;
+  if (Array.isArray(grpcServiceConfig.methodConfig)) {
```
sofisl, line 395: Should there be an error or alert if methodConfig is not an array?

typescript/src/schema/proto.ts
```diff
@@ -393,22 +392,26 @@ function getMethodConfig(
...
+    for (const config of grpcServiceConfig.methodConfig) {
+      if (!config.name) {
+        continue;
       }
-      if (name.service === serviceName && name.method === methodName) {
-        exactMatch = config;
+      for (const name of config.name) {
+        if (name.service === serviceName && !name.method) {
```
sofisl, line 401: Seems kind of strange to me that this wouldn't be an if/else, because would it be possible for !name.method and name.method ==== methodName?

typescript/test/test-application/test-ts.ts
```diff
@@ -67,18 +68,18 @@ describe('Test application for TypeScript users', () => {
     this.timeout(240000);
     // copy protos to generated client library and copy test application to local.
     if (!fs.existsSync(path.join(showcaseLib, 'protos'))) {
-      fs.copySync(protos, path.join(showcaseLib, 'protos'));
+      await copy(protos, path.join(showcaseLib, 'protos'));
```
sofisl: Maybe these tests could check the existence of the files after copying?

alexander-fenster: Oh. So the thing is, in the recent commit, I removed "test applications" because they were just too flaky and were actually broken for some time. So this code is gone completely.

typescript/src/schema/proto.ts
alexander-fenster: Yes, this is a good idea to let the user know that their gRPC service config is broken. Added a warning. (failure might be too much I guess?)

typescript/src/schema/proto.ts
alexander-fenster: Agreed, if-else is more logical here. Fixed.

package.json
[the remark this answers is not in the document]
sofisl: Will this mean we need to update our scripts in generated libraries? i.e.: https://github.com/googleapis/google-cloud-node/blob/48763577ffe3a47a445c3c3f53a4801049f5d217/packages/google-appengine/package.json#L40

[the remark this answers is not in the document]
sofisl: With this in mind, might be worth testing this version of the generator with a current library

sofisl (approved):

package.json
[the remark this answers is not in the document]
alexander-fenster: No need to update scripts in the generated library, it's just for the generator in its own `package.json`. Having said that, I did test libraries.

WORKSPACE
```diff
@@ -18,10 +13,34 @@ load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies", "rules_
...
+    ts_version_from = "//:package.json",
+)
 
-yarn_install(
+# Fetch and register node, if you haven't already
+load("@rules_nodejs//nodejs:repositories.bzl", "nodejs_register_toolchains")
+nodejs_register_toolchains(
+    name = "nodejs",
+    # This is the version of Node.js that would run the generator, it's unrelated to the versions supported by the generated libraries
+    node_version = "18.12.1", # https://github.com/bazelbuild/rules_nodejs/blob/stable/nodejs/private/node_versions.bzl
```
noahdietz, line 29: Nit: can/should this be referenced from a single spot such that when you want to update this value, the other spots (e.g. CI, package.json or whatever) are also updated?

typescript/test/unit-test-runner.ts
```diff
@@ -0,0 +1,33 @@
...
+  const files = await fsp.readdir(testDirectory);
+  const tests = files.filter(file => file.endsWith('.js'));
+  for (const test of tests) {
+    mocha.addFile(path.join(process.cwd(), testDirectory, test));
+  }
+  await mocha.loadFilesAsync();
+  mocha.run((failures: number) => { process.exitCode = failures; });
+}
+
+main();
```
noahdietz: Nit: add a new line at the end of the file?

noahdietz (commented):
LGTM for bazel stuff

typescript/test/unit-test-runner.ts
alexander-fenster: It's there now!

WORKSPACE
alexander-fenster: Added a constant in `repositories.bzl` and imporing it from the workspace file. The same import will be added in the googleapis `WORKSPACE`.

bcoe (approved):

PR #1328, opened 2023-02-13
fix: use correct syntax for inline link tags

hegemonic:
Google Cloud uses [API Extractor](https://api-extractor.com/) and [API Documenter](https://api-extractor.com/pages/setup/generating_docs/) to generate API docs for Node.js client libraries. These tools can't resolve links that are formatted like the following example:

```
[Widget]{@link google.someapi.v1.Widget}
```

However, links that use [TSDoc notation](https://api-extractor.com/pages/tsdoc/tag_link/) work just fine:

```
{@link google.someapi.v1.Widget | Widget}
```

So I've updated the templates to use the latter syntax.

@dansaadati FYI


alexander-fenster (approved):
Thank you!

PR #1331, opened 2023-02-15
fix: update generated samples to reflect string arrays

sofisl:
fixes: b/232825963


templates/typescript_samples/samples/generated/$version/_util.njk
```diff
@@ -86,12 +86,14 @@
             'Buffer.from('string')'
```
alexander-fenster, line 86: Unrelated to this PR, but this does not look right, probably the outer apostrophes are not needed?

alexander-fenster (approved):

PR #1343, opened 2023-02-21
fix: rename snippet metadata file

sofisl:
Fixes b/256852641


alexander-fenster:
If you are renaming it, do you need to `git rm` the old files from the baselines?

alexander-fenster (approved):

PR #1351, opened 2023-03-24
fix: ensure links are prefixed with "protos."

dansaadati:
This will fix links to protos, which are currently unresolvable without this prefix.


alexander-fenster (approved):

alexander-fenster:
Thank you! I'll release this on Monday.

PR #1391, opened 2023-05-30
fix: do not generate docs-test script

alexander-fenster:
Since we stopped running `docs-test` script as a part of our CI in `google-cloud-node`, let's stop pulling `linkinator` as a dev dependency, and stop adding the `docs-test` script for the new libraries.

This change will only affect newly generated libraries.


bcoe (approved):
LGTM, although I would like us to figure out another place where we can catch broken links.

Early bird?

sofisl (approved):

PR #1402, opened 2023-06-15
feat: add ESM path to generator

sofisl:
This PR adds a path to generate ESM-templates in the generator by adding a `format` command-line variable. If omitted, it will generate (regular) cjs code. If included, it will generate ESM-code, that transpiles into dual format. Templates now live under `cjs` and `esm` directories.


sofisl:
@ddelgrosso1, these are the changes for the generator. This is failing because it's attempting to run `tsc -p ./tsconfig.json`, so hopefully once we get your plug-in solution this will be resolved!

WORKSPACE
```diff
@@ -40,4 +40,4 @@ npm_translate_lock(
 
 load("@npm//:repositories.bzl", "npm_repositories")
 npm_repositories()
-pnpm_repository(name = "pnpm")
+pnpm_repository(name = "pnpm")
```
alexander-fenster, line 43: I believe this file can be reverted since the only change is the linebreak.

baselines/asset/package.json
```diff
@@ -45,8 +45,8 @@
     "gapic-tools": "^0.1.8",
     "gts": "5.0.1",
     "jsdoc": "^4.0.2",
-    "jsdoc-region-tag": "^3.0.0",
     "jsdoc-fresh": "^3.0.0",
+    "jsdoc-region-tag": "^2.0.1",
```
alexander-fenster, line 49: Why the downgrade?

protos/index.d.ts
```diff
@@ -1,7 +1,5 @@
 import * as $protobuf from "protobufjs";
-
-// Note: this file was manually edited to remove references to Long which are not needed in this project.
```
alexander-fenster, line 3: I think it's a good idea to remove references to `long`.

typescript/src/schema/proto.ts
```diff
@@ -26,7 +26,17 @@ import {BundleConfig} from '../bundle.js';
 import {Options} from './naming.js';
 import {ServiceYaml} from '../serviceyaml.js';
 import protobuf from 'protobufjs';
-import protoJson from '../../../protos/protos.json' assert {type: 'json'};
```
alexander-fenster, line 29: Why is this change needed? What's wrong with the regular import? The path is static anyway, isn't it?

typescript/test/unit/util.ts
```diff
@@ -25,6 +25,11 @@ import protobuf from 'protobufjs';
 import protoJson from '../../../protos/protos.json' assert {type: 'json'};
 import type * as protos from '../../../protos/index.js';
 
+import * as fs from 'fs';
```
alexander-fenster, line 28: Please remove unused imports.

typescript/test/util.ts
```diff
@@ -137,12 +138,17 @@ export function runBaselineTest(options: BaselineOptions) {
...
+    if (options.format) {
+      commandLine += ` --format="${options.format}"`;
+    }
     if (options.diregapic) {
       commandLine += ' --diregapic';
     }
     if (options.mixins) {
       commandLine += ` --mixins="${options.mixins}"`;
     }
+    console.log('COMMANDLINE ARG:');
```
alexander-fenster, line 150: Please remove the extra logging.

typescript/tools/update-baselines.ts
```diff
@@ -114,7 +114,11 @@ async function main() {
 
     console.log(`Updating baseline for ${library}...`);
     console.log(`  - rm -rf "${baselineDir}"...`);
-    await fsp.rm(baselineDir, {recursive: true});
+    try {
+      await fsp.rm(baselineDir, {recursive: true});
+    } catch (err) {
+      console.log('no baseline to delete');
```
alexander-fenster, line 120: If we just ignore this, let's ignore silently. Otherwise, this log messages is not very helpful, should've said "Not removing baseline `${baselineDir}` because it does not exist", or something like that.

alexander-fenster (approved):

baselines/asset/package.json
sofisl: Good catch!

PR #1441, opened 2023-08-31
chore: re-introduce gts and fix all linting issues

### LINKED ISSUE #560: bring back gts and/or eslint
As a part of moving to ESM module syntax, I disabled `gts` / `eslint` (partially because it does not understand the new `assert { type: 'json' }` import syntax). We need to bring the code formatting back.

ddelgrosso1:
fixes: googleapis/google-cloud-node-core#560

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,7 @@
+**/node_modules
+**/coverage
+test/fixtures
+build/
+docs/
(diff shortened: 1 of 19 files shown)


alexander-fenster (approved):
Thank you!

PR #1511, opened 2024-01-04
feat: add UUID autogeneration for fields listed as autoPopulate in service.yaml

sofisl:


sofisl:
@alexander-fenster, I think the failing tests are due to gapic-tools not recognizing the field_info.proto. Once we publish a new version of gax with the new version of nodejs-proto-files, I can also update gapic-tools and I think it will just work.

alexander-fenster:
Let me review this after #1503 gets in.

baselines/asset/src/v1/asset_service_client.ts.baseline
```diff
@@ -22,6 +22,7 @@ import type {Callback, CallOptions, Descriptors, ClientOptions, GrpcClientOption
 
 import * as protos from '../../protos/protos';
 import jsonProtos = require('../../protos/protos.json');
+import * as crypto from 'crypto';
```
alexander-fenster, line 25: This immediately disqualifies the library from the browser or any other non-Node.js usage. One way to solve this would be to move everything that needs `crypto` to google-gax and export it from there, and in gax, make it similar to https://github.com/googleapis/google-auth-library-nodejs/tree/main/src/crypto to use either Node.js crypto or `window.crypto.subtle`, whichever is available.

Also, I don't see any `crypto` usage in this particular file, which probably means that this import should be conditional.

templates/esm/typescript_gapic/esm/src/$version/$service_client.ts.njk
```diff
@@ -607,6 +608,9 @@ export class {{ service.name }}Client {
 {%- endif %}
       ]>|void {
     request = request || {};
+    {% for field in method.autoPopulatedFields %}if (!request.{{ field.toCamelCase() }}) {
+      request.{{ field.toCamelCase() }} = crypto.randomUUID();
```
alexander-fenster, line 612: It does not seem like this is covered in any baseline tests - we should have a baseline that would contain this line.

alexander-fenster, line 612: To make it compatible with browser, we have two options: either use `uuid` module instead of crypto (which will force us make `uuid` a dependency for all libraries), or - probably better - export a `createUUID()` function from gax, and have it there instead.

typescript/src/schema/proto.ts
```diff
@@ -74,6 +75,7 @@ export interface MethodDescriptorProto
 
 export interface ServiceDescriptorProto
   extends protos.google.protobuf.IServiceDescriptorProto {
+  getAutoPopulatedFields: MethodDescriptorProto[];
```
alexander-fenster, line 78: Is this field actually used anywhere?  I see the function `getAutoPopulatedFields` but I don't see if this added field is accessed anywhere.

baselines/showcase-esm/esm/src/v1beta1/echo_client.ts.baseline
```diff
@@ -459,6 +461,9 @@ export class EchoClient {
         protos.google.showcase.v1beta1.IEchoRequest|undefined, {}|undefined
       ]>|void {
     request = request || {};
+    if (!request.requestId) {
+      request.requestId = gax.makeUUID();
```
alexander-fenster, line 465: Could we have a generated unit test that would check that this code is actually executed? It would be similar to an existing test that makes a request, but in a stub, it can check that the field was indeed assigned to some UUID.  Similar to https://github.com/googleapis/gapic-generator-typescript/blob/2a61552f77d302f15bccabfaded698f1e2b07e12/templates/esm/typescript_gapic/esm/test/gapic_%24service_%24version.ts.njk#L319.

typescript/src/schema/proto.ts
sofisl: REmoved!

alexander-fenster (approved):

PR #1558, opened 2024-03-14
feat: support GOOGLE_CLOUD_UNIVERSE_DOMAIN environment variable

alexander-fenster:
According to the spec, this environment variable points to the universe if it's not configured in the code.  If it's set in the code, the value from the code has priority.

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -1,5 +1,5 @@
 {
-  "lockFileVersion": 3,
+  "lockFileVersion": 6,
   "moduleFileHash": "0e3e315145ac7ee7a4e0ac825e1c5e03c068ec1254dd42c3caaecb27e921dc4d",
   "flags": {
     "cmdRegistries": [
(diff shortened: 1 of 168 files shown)


alexander-fenster:
@sofisl Can you take a look what's going on with the ESM tests?

```
node_modules/@js-sdsl/ordered-map/dist/esm/index.d.ts:402:35 - error TS2748: Cannot access ambient const enums when 'isolatedModules' is enabled.

402 export type { OrderedMapIterator, IteratorType, Container, ContainerIterator, TreeContainer };
                                      ~~~~~~~~~~~~
```

Is it caused by some dependency update?

sofisl:
> Is it caused by some dependency update?

it is caused by the @js-sdl dep being added to grpc: https://github.com/grpc/grpc-node/pull/2677/files

We can just turn off `isolatedModules` flag, it won't affect the generated code and, since it's in CJS, none of our other libraries use it.

sofisl (approved):

PR #1571, opened 2024-04-08
feat: use ClientOptions gaxServerStreamingRetries value

leahecole:
Instead of setting gaxServerStreamingRetries to false, set it to the value in ClientOptions or default to false if not set

Tests in the gax-nodejs test-application should be passing in https://github.com/googleapis/gax-nodejs/pull/1583


leahecole:
Ooo, there are clearly some other things I need to update - I have to run to a meeting but I will read the error logs to see what needs to be changed. If it's not obvious though, any hints of what I need to fix would be appreciated!

alexander-fenster:
When you update templates, you also need to update baselines; `npm run baseline` will do the trick!  (takes a few minutes, give it some time)

sofisl (approved):

PR #1575, opened 2024-04-15
fix: do not use gaxServerStreamingRetries for bidi or client streaming

leahecole:
In preparation for selectively opting client libraries into the change, this PR adds some more fine grained changes for gax server streaming retries:

- only passes the gaxServerStreamingRetries opt-in flag for server streaming calls (not for bidi and client streaming)
- adds happy path generated test cases

I have tested this successfully on nodejs-bigquery-storage and am testing it now with a new regeneration of the showcase test application as part of https://github.com/googleapis/gax-nodejs/pull/1587


sofisl (approved):
LGTM, small nit, please change title to fix: do not use gaxServerStreamingRetries for bidi or client streaming

PR #1576, opened 2024-04-19
feat: add apiVersion to x-goog headers

sofisl:
see https://docs.google.com/document/d/1ZOrybNB1BUFCFzykqzLuhwL9HHqn0fJV6ZizhVEROxs/edit?tab=t.0

I've manually regenerated showcase and confirmed the headers are sent correctly, as described in the doc above.

### DIFF
--- a/templates/cjs/typescript_gapic/src/$version/$service_client.ts.njk
+++ b/templates/cjs/typescript_gapic/src/$version/$service_client.ts.njk
@@ -249,7 +249,7 @@ export class {{ service.name }}Client {
     }
 {%- endif %}
 {%- if service.apiVersion %}
-    clientHeader.push({{ service.apiVersion }});
+    clientHeader.push('x-goog-api-version': {{ service.apiVersion }});
 {%- endif %}
     if (opts.libName && opts.libVersion) {
       clientHeader.push(`${opts.libName}/${opts.libVersion}`);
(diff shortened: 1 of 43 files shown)


templates/cjs/typescript_gapic/src/$version/$service_client.ts.njk
```diff
@@ -389,7 +389,10 @@ export class {{ service.name }}Client {
     // Put together the default options sent with requests.
     this._defaults = this._gaxGrpc.constructSettings(
         '{{ api.naming.protoPackage }}.{{ service.name }}', gapicConfig as gax.ClientConfig,
-        opts.clientConfig || {}, {'x-goog-api-client': clientHeader.join(' ')});
+        opts.clientConfig || {}, {'x-goog-api-client': clientHeader.join(' ')
+{%- if service.apiVersion -%}
+       , 'x-goog-api-version': '{{ service.apiVersion }}'
```
alexander-fenster, line 394: It seems like this change is not covered by baselines, so if it stops working for whatever reason, we won't be able to catch it. Should we add the corresponding annotation to at least one baseline service?

sofisl:
I've added the updated protos from here: https://github.com/googleapis/gax-nodejs/blob/main/gax/test/showcase-echo-client/protos/google/showcase/v1beta1/echo.proto

So the generated code includes lots of changes that have already been included in the generator, but weren't updated due to old protos.

alexander-fenster (approved):

PR #1661, opened 2024-10-18
feat: paging changes for bigquery

leahecole:
This PR resolves b/360115234 - see that bug and its parent b/352331075 for initial proposal info. Things this PR does include:
- the addition of a boolean field in the `MethodDescriptorProto` interface called `maxResultsParameter` - this is true if integer wrapper types are allowed and a method has a `maxResults` parameter
- logic in the templates that when a method has `maxResultsParameter` set to true, if it is a paginated method (but not async or stream), convert a `number` type that the user passes in and reformat it to be in the proper wrapped integer type the underlying API expects
- logic in the templates that when a method has `maxResultsParameter` set to true, if it is a paginated method (but not an async or stream one), and if the parameter is `maxResults`, update the param comments, allow either the wrapper type or the `number` type
-logic in the samples templates that when a method has `maxResultsParameter` set to true, if it is a paginated method (but not an async or stream one), and if the parameter is `maxResults`, update the TODO comments to have the example comment be a number type instead of the protobuf one
- adds new baselines for bigquery v2
    - adds protos in test-fixtures
    - updates ci.yaml to call baselines
    - updates baseline-esm.ts and cjs to run

cc @shollyman @alvarowolfx


leahecole:
Noted! Re point 3 I think that I redid the baselines before I took out a comment I had in a nunjucks file and so that makes sense why it hit so many. I will work on a few outstanding changes, make those baselines, and get this ready for real review

leahecole:
I added baselines @sofisl! I am honestly not sure why the pubsub baseline regenerates. However, it regenerated with the correct behavior - that maxResultsParameter is set to false because pubsub isn't an allowlisted service, so I am less concerned than if it regenerated with maxResultsParameter set to true.

sofisl:
I think we'll also want to run regular + esm tests [here](https://github.com/googleapis/gapic-generator-typescript/blob/0d5ca4a1d9c7dec5628de507e1715701f3fbf7f6/.github/workflows/ci.yaml#L77)

sofisl (changes requested):
I think we'll also want to add cjs+ esm tests [here](https://github.com/googleapis/gapic-generator-typescript/blob/0d5ca4a1d9c7dec5628de507e1715701f3fbf7f6/.github/workflows/ci.yaml#L77) and make sure they pass

typescript/src/schema/proto.ts
```diff
@@ -491,6 +563,11 @@ function augmentMethod(
   parameters: AugmentMethodParameters,
   method: MethodDescriptorProto
 ) {
+  // whether a service is allowed to use Int32Value and UInt32Value wrappers - generally this is only BigQuery
+  // this is used to determine factors about pagination fields and to allow users to pass a "number" instead of
+  // having to convert to a protobuf wrapper type to determine page size
+  const wrappersAllowed =
+    ENABLE_WRAPPER_TYPES_FOR_PAGE_SIZE[parameters.service.packageName];
```
sofisl, line 570: I think if you don't want pubsub to change, you'd make this:
```ENABLE_WRAPPER_TYPES_FOR_PAGE_SIZE[parameters.service.packageName] ? true : undefined```
This would preserve current behavior but it would also make it less clear in the future. However, do you envision this list being actually a full dict of opt-in or opt-outs? I could also see `ENABLE_WRAPPER_TYPES_FOR_PAGE_SIZE` just being an array where we assume opt-in. I'm fine either way, perhaps it's best to talk to pubsub since it seems they're the only other ones potentially affected by this change (i.e., do they want to explicitly opt-out or just continue functionality?)

Comment (leahecole) [reply typescript/src/schema/proto.ts:570]: @feywind thoughts? I'll set up time to have a quick chat to see if this would help or hurt things on your end

Comment (leahecole) [reply typescript/src/schema/proto.ts:570]: RE whether it should be opt-in, I think it should be opt-ins

Comment (feywind) [reply typescript/src/schema/proto.ts:570]: Yeah, let's maybe chat about it, but it sounds like it's probably not going to be an issue, as long as it doesn't require a major bump in Pub/Sub.

**Comment by leahecole** [comment]
@googleapis/jsteam-handwritten-libraries FYI, this is part of some modernization work being done for BQ - if your library is also impacted by this, feel free to ping me and I can share more

File: templates/cjs/typescript_gapic/src/$version/$service_client.ts.njk [hunk 2]
```diff
@@ -893,10 +894,23 @@ export class {{ service.name }}Client {
         {{ util.toInterface(method.outputInterface) }}
       ]>|void {
     request = request || {};
+  {%- if method.maxResultsParameter %}
+    // the underlying API expects a UInt32Value or Int32Value object
```
sofisl, line 898: I would actually get rid of this comment or perhaps reword it, as technically this is what goes in the client library (and thus what users actually use). Maybe just say:

```
Converts number to Unit32 or Int32 value for non-compliant APIs.
```

typescript/src/schema/proto.ts
```diff
@@ -35,6 +35,12 @@ const COMMON_PROTO_LIST = [
   'google.type',
 ];
 
+// services that are allowed to use Int32Value and UInt32Value protobuf wrapper types
+// instead of "number" for pageSize/maxResults
+// keyed by proto package name, e.g. "google.cloud.foo.v1".
+const ENABLE_WRAPPER_TYPES_FOR_PAGE_SIZE = {
```
sofisl, line 41: nit, this could be simplified if we just assume this list is opt-in, and remove the `true`flag, but up to you!

sofisl (approved):

templates/cjs/typescript_gapic/src/$version/$service_client.ts.njk
leahecole: Done

PR #1673, opened 2025-01-23
fix: typo in page stream comment template

leahecole:
I think this was just a copy paste error that has persisted in the templates for awhile! 
In case it has trouble loading because of the baselines, this is it
image


leahecole:
nvm, we have determined this actually was right before, I misunderstood what the comment was saying 🤦🏻

PR #1677, opened 2025-01-31
fix: add type checking for headers

sofisl:
Does not include tests - see googleapis/google-cloud-node-core#549

### DIFF
--- a/baselines/redis-esm/esm/src/v1beta1/cloud_redis_client.ts.baseline
+++ b/baselines/redis-esm/esm/src/v1beta1/cloud_redis_client.ts.baseline
@@ -1334,7 +1334,7 @@ export class CloudRedisClient {
    * @return {Promise} - The promise which resolves to an array.
    *   The first element of the array is an object representing
    * {@link google.longrunning.Operation | google.longrunning.Operation}.
-   * The promise has a method named "cancel" which cancels the ongoing API call.
+   * The promise has a methodƒC named "cancel" which cancels the ongoing API call.
    *
    * @example
    * ```
@@ -1359,12 +1359,12 @@ export class CloudRedisClient {
       {} | null | undefined
     >
   ): Promise<[protos.google.longrunning.Operation]> {
-     let options: CallOptions;
+     let options: gax.CallOptions;
      if (typeof optionsOrCallback === 'function' && callback === undefined) {
        callback = optionsOrCallback;
        options = {};
      } else {
-       options = optionsOrCallback as CallOptions;
+       options = optionsOrCallback as gax.CallOptions;
      }
      options = options || {};
      options.otherArgs = options.otherArgs || {};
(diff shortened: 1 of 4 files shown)


leahecole (approved):

baselines/redis-esm/esm/src/v1beta1/cloud_redis_client.ts.baseline
```diff
@@ -1360,12 +1360,12 @@ export class CloudRedisClient {
     >
   ): Promise<[protos.google.longrunning.Operation]> {
     options = options || {};
```
d-goog, line 1362: Instead of casting below, would it be possible to declare the type here?

```suggestion
    options: gax.CallOptions = options || {};
```

sofisl: This is where my TS knowledge gets murky - since options could be a Callback type as well, would casting be better since it won't change the underlying type, just say "read it as if it were a gax.CallOption" vs. "expect a gax.CallOption"? I actually don't know, I'll defer to you here.

d-goog: Np, we could use this to narrow the types:
```suggestion
    options: gax.CallOptions = typeof options === 'object' ? options  : {};
```

Casting the function technically works (since functions are technically objects, while not `typeof` objects), but we may not want to add properties to them (unless something downstream is expecting it)

sofisl: technically the API is expecting the header. So here's my concern with your suggesting, callback can be:

```
      | Callback<
          protos.google.protobuf.Empty,
          protos.google.longrunning.CancelOperationRequest,
          {} | undefined | null
        >
```
so I worry casting `options: gax.CallOptions = typeof options === 'object' ? options  : {};` in the second case, might not actually be true to Callback. 

I wonder if we could do some typechecking? Something like `if options.otherArgs then typeof options else  typeof       Callback<protos.google.protobuf.Empty,protos.google.longrunning.CancelOperationRequest, | undefined | null> but then I do think that regardless we need to add this header arg, or else the methods don't work. WDYT?

sofisl: Right, but the falsy condition of the ternary operation would make it an empty object, which I don't think type callback is. Although I'm realizing this is an issue with the line of code regardless, which I copied over from other LRO operation polling. Let me check what the potential types are there.

sofisl: Ok, added some code from what the other LRO operations do.

d-goog (approved):

PR #1692, opened 2025-03-10
fix: add json files to tsconfig templates

sofisl:


templates/cjs/typescript_gapic/tsconfig.json.njk
```diff
@@ -28,6 +28,8 @@ limitations under the License.
     "src/**/*.ts",
     "test/*.ts",
     "test/**/*.ts",
-    "system-test/*.ts"
+    "system-test/*.ts",
+    "src/**/*.json",
+    "protos/protos.json"
```
leahecole, line 33:
```suggestion
    "samples/**/*.json",
    "protos/protos.json"
```

templates/esm/typescript_gapic/tsconfig.esm.json.njk
```diff
@@ -36,9 +36,8 @@ limitations under the License.
     "esm/src/**/*.ts",
     "esm/test/*.ts",
     "esm/test/**/*.ts",
-    "esm/system-test/*.ts"
-  ],
-  "exclude": [
-    "esm/src/json-helper.cjs"
+    "esm/system-test/*.ts",
+    "esm/src/**/*.json",
```
leahecole, line 40:
```suggestion
    "esm/src/**/*.json",
    "esm/samples/**/*.json",

```

leahecole (changes requested):
I think we need this for samples too! Or at least we do in BQ storage :)

templates/cjs/typescript_gapic/tsconfig.json.njk
sofisl: Samples shouldn't be included in ts linting, so I think better to add samples to eslintignore (if it's not already there!)

leahecole: Duh, makes sense!

feywind (approved):

leahecole (approved):

PR #1697, opened 2025-03-14
fix: remove floating promises

sofisl:
Fixes issue with Node 18 that linting has been blocked due to un-awaited promises. This should fix it.


sofisl:
See #1698

danieljbruce (approved):
Looks fine to me.

The only line that concerned me was opting out of gax retries at:
https://github.com/googleapis/gapic-generator-typescript/pull/1697/files#diff-d1e77a3ac48fccec73f27d57cdef9c62f6cd7173640fa6159e7d00da6bc57c6aR618

But it should be good to go if this is really just a red herring.

danieljbruce:
The link I posted doesn't work. I mean the line that says this:

`const client = new {{ service.name.toLowerCase() }}Module.{{ api.naming.version }}.{{ service.name }}Client({{- util.initClientOptions(api.rest, true)`

PR #1715, opened 2025-04-02
chore: add selective gapic unit tests

kevkim-codes:
Makes sure service yaml parsing is happening properly in the `augmentService` from the `proto.ts` file.


typescript/test/unit/proto.ts
```diff
@@ -1407,4 +1409,317 @@ describe('src/schema/proto.ts', () => {
...
+            paramType: 'TYPE_STRING',
+            comments: [' A random request_id to test autopopulation'],
+            fieldBehavior: 1,
+            fieldInfo: {format: 1},
+          } as Comment;
+        },
+      };
+
+      const options = {
+        serviceYaml: {
```
danieljbruce, line 1680: nit: According to the description I guess we are checking that the parsing of this is correct. If so then maybe an assert statement here?

danieljbruce (approved):
Looks fine as is. A few optional suggestions, but this PR is just adding tests so definitely fine if we just merge this PR as is.

typescript/test/unit/proto.ts
```diff
@@ -835,18 +837,18 @@ describe('src/schema/proto.ts', () => {
...
           },
-          getServiceComment: function (serviceName: string): string[] {
+          getServiceComment: function (_serviceName: string): string[] {
             return ['not needed'];
           },
           getMethodComments: function (
-            serviceName: string,
-            methodName: string
+            _serviceName: string,
+            _methodName: string
```
danieljbruce, line 845: nit: What is the reason we are prefixing this parameter name with an underscore now?

kevkim-codes: It's prettier convention for unused param names. I didn't want to just delete them from the unit tests (assuming they were there originally for a reason).

typescript/test/unit/proto.ts
kevkim-codes: The asserts at the end of the unit tests assess if the parsing happened correctly. If the method exists in the service or does not, then we can tell if the parsing was successful or not.

PR #1755, opened 2025-05-13
fix: Use the new cached method fromJSON from google-gax

mkruskal-google:


sofisl:
would you mind running `npm run baseline`? It

sofisl (approved):

noelportillo (commented):
Good

PR #1758, opened 2025-05-27
feat: add pb/version in client header.

miguelvelezsa:
Add protobufjs version in client headers. Needed to track protobufjs version used by clients.


sofisl:
This is obsolete now! Sorry and thank you for your work :)

PR #1846, opened 2025-09-30
fix: broken compileProto step and add tests

### LINKED ISSUE #539: Process: add tests for api_naming and for the bazel rule
we could have added more tests here:

https://github.com/googleapis/gapic-generator-typescript/pull/1837

specifically to check that the naming looks right, and ideally (in the future) for the typescript_gapic_combined_pkg.bzl. Even though it is a bash script, I bet we could test it a bit more.

### LINKED ISSUE #1839: Process: add tests for api_naming and for the bazel rule
we could have added more tests here:

https://github.com/googleapis/gapic-generator-typescript/pull/1837

specifically to check that the naming looks right, and ideally (in the future) for the typescript_gapic_combined_pkg.bzl. Even though it is a bash script, I bet we could test it a bit more.

sofisl:
Unfortunately this PR: https://github.com/googleapis/google-cloud-node/pull/6699 broke us without realizing it. The fix is here: https://github.com/googleapis/google-cloud-node/pull/6766, but I've also added a test to this library since we would have caught the breakage when we went to upgrade the library  (https://github.com/googleapis/gapic-generator-typescript/pull/1833). It took me a lot of time to debug it, so I figured a test was finally due.

This PR looks big, but only because I added a lot of test-fixtures files. The only real change was that I separated out the script that used to be in rules_typescript_gapic/typescript_gapic_combined_pkg.bzl into its own file (rules_typescript_gapic/combine_script.sh) so that I could test it on its own. Then I added a test in .github/workflows/ci.yaml.

It also does one small fix for calling the compileProtos binary (it's really a typo, instead of being called with quotes around the variable name, which causes a whitespace to be passed and caused it to silently fail, I'm calling the variables without a whitespace). Again, this test should cover it, since we are generating a library and running tests in the library with the output of the test.

Partially fixes googleapis/google-cloud-node-core#539


package.json
```diff
@@ -63,7 +63,7 @@
     "@types/object-hash": "^3.0.6",
     "@types/yargs": "^17.0.33",
     "espower-typescript": "^10.0.1",
-    "gapic-node-processing": "^0.1.2",
+    "gapic-node-processing": "^0.1.21-experimental",
```
leahecole, line 66: Will we be able to get automated dependency updates to this? If not, is there a tracking bug for updating this once it's not experimental?

leahecole (approved):

PR #1853, opened 2025-10-01
fix: update gapic-node-processing and add ESM tests

sofisl:
Since we have to update gapic-node-processing here anyways: https://github.com/googleapis/google-cloud-node/pull/6768/files

I figured I'd add a test and make sure we'd catch it at the generator-level as well.


.github/workflows/ci.yaml
```diff
@@ -179,6 +179,35 @@ jobs:
         npm test
         npm run system-test
 
+  combined-lib-test-esm:
+    needs: build
+    runs-on: ubuntu-latest
+    container: gcr.io/gapic-images/googleapis:20250404
```
leahecole, line 185: I see this container has a date - will it get automatically updated by our tools?

.github/workflows/ci.yaml
```diff
@@ -179,6 +179,35 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
```
leahecole, line 187: If this is v5, why does it have a SHA specified rather than a version number?

package.json
```diff
@@ -63,7 +63,7 @@
     "@types/object-hash": "^3.0.6",
     "@types/yargs": "^17.0.33",
     "espower-typescript": "^10.0.1",
-    "gapic-node-processing": "^0.1.3",
+    "gapic-node-processing": "^0.1.22-experimental",
```
leahecole, line 66: same note from the other PR about changing this :)

leahecole (approved):

PR #1894, opened 2025-11-05
fix: add per-service apiVersion to gapic_metadata

noahdietz:
Populates the `api_version` field on a per-`service` basis to the generated `gapic_metadata.json` using the value from the `service`-level annotation `google.api.api_version` already parsed for use in the `X-Goog-Api-Version`/`$apiVersion` system parameter (as of #1576).

Showcase-based tests already have `google.api.api_version` set on `service Echo`, so the baselines are appropriately updated to account for inclusion of `apiVersion` in the `gapic_metadata.json` output.

It's presence in `gapic_metadata.json` will allow parsers to correlate `service` client to API version. It is not in use by any API yet. Only the unit tests needed updating.

Internal bug http://b/452365074.

### DIFF
--- a/baselines/showcase-esm/esm/src/v1beta1/gapic_metadata.json.baseline
+++ b/baselines/showcase-esm/esm/src/v1beta1/gapic_metadata.json.baseline
@@ -120,6 +120,7 @@
       }
     },
     "Echo": {
+      "apiVersion": "v1_20240408",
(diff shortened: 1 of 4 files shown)


noahdietz:
Closing in favor of https://github.com/googleapis/google-cloud-node-core/pull/788

